PR #4, opened 2015-09-03
Switch from "wget" to "curl"

tianon:
This avoids strange intermittent issues like:
```
Connecting to test.docker.com (216.137.36.117:443)
wget: error getting response: Connection reset by peer
```


tianon:
Yay, fixes the build! :+1:

PR #7, opened 2016-01-06
Update to alpine:3.3, especially for --no-cache on "apk add"

tianon:


tianon:
Those Travis build/test times -- 44s and 45s.  :smile:

yosifkit:
LGTM

PR #13, opened 2016-06-30
Pre-add "dockremap" user and group with subuid/subgid info so that "--userns-remap=default" works out of the box

tianon:


yosifkit:
LGTM

PR #16, opened 2016-08-02
brackets added to DOCKER_VERSION env var

jojow:
so it is consistent with other env vars


tianon:
Not strictly necessary, but more consistent and a fair point, thanks! :+1:

PR #24, opened 2016-09-02
Added support "bash" variant

SuperPaintman:
Hi! I added "bash" variant

### DIFF
--- a/1.11/bash/Dockerfile
+++ b/1.11/bash/Dockerfile
@@ -1,0 +1,6 @@
+FROM docker:1.11
+
+RUN apk add --no-cache \
+    bash
+
+CMD ["bash"]
--- a/1.11/bash/Dockerfile
+++ b/1.11/bash/Dockerfile
@@ -1,6 +1,6 @@
 FROM docker:1.11
 
 RUN apk add --no-cache \
-    bash
+		bash
 
 CMD ["bash"]
--- a/1.12/bash/Dockerfile
+++ b/1.12/bash/Dockerfile
@@ -1,0 +1,6 @@
+FROM docker:1.12
+
+RUN apk add --no-cache \
+		bash
+
+CMD ["bash"]
--- a/1.12/experimental/bash/Dockerfile
+++ b/1.12/experimental/bash/Dockerfile
@@ -1,0 +1,6 @@
+FROM docker:1.12-experimental
+
+RUN apk add --no-cache \
+    bash
+
+CMD ["bash"]
--- a/1.12/experimental/bash/Dockerfile
+++ b/1.12/experimental/bash/Dockerfile
@@ -1,6 +1,6 @@
 FROM docker:1.12-experimental
 
 RUN apk add --no-cache \
-    bash
+		bash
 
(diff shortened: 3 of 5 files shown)


tianon:
I'm not sure what the intended use-case here is -- can you elaborate?  We have a `git` variant only because the client actually supports using `git`, but its use is rare enough not to be necessary for most users (and the size difference is significant enough to make it worthwhile to do so).

SuperPaintman:
This isn't question of `git`.

`sh` != `bash`

## For example

**docker:git**, **docker** - images:

```sh
#!/bin/sh

# №1
#======

arr=()
# STDERR => sh: syntax error: unexpected "("


# №2
#======
function hello { echo "github"; }
# STDERR => sh: syntax error: unexpected "}"
```


**superpaintman/docker:bash** image:

```sh
#!/bin/bash

# №1
#======

arr=()
arr+=("hello")
arr+=("superpaintman/docker")

echo "${arr[@]}"
# STDOUT => hello superpaintman/docker


# №2
#======
function hello { echo "github"; }

hello
# STDOUT => github
```

[and other differences](//mywiki.wooledge.org/Bashism).


## About use-cases:

* [Gitlab CI](//docs.gitlab.com/ce/ci/yaml/README.html)

```yaml
# ...

deploy:
    stage: deploy
    script: |
        arr=()

        if [ "${all_files+1}" ]; then
            arr+=("file1.txt")
        fi

        arr+=("file2.txt")

        for a in ${arr[@]}; do
            scp ./$a gitlab-ci@192.168.1.100:/tmp/
        done
```


* Running `bash` scripts
* Using bash lang-constructions
* Etc

tianon:
I realize that Bash isn't `/bin/sh` -- my point was that the `git` variant exists to support core Docker features directly which cannot work without `git` installed.  To my knowledge, there exist no such features which use Bash directly within Docker.

What I would recommend is using a simple `Dockerfile` to add the customization you require, especially and including the scripts you plan to run:
```dockerfile
FROM docker:1.12
RUN apk add --no-cache bash
COPY my-script.sh /usr/local/bin/
CMD ["my-script.sh"]
```

Combine that with automated builds (https://docs.docker.com/docker-hub/builds/) and repository links (https://docs.docker.com/docker-hub/builds/#repository-links) and it's reasonably easy to have an up-to-date image built `FROM` this one with custom features such as Bash and user-supplied scripts included by default.

SuperPaintman:
Oh, ok. I didn't think that these images are made for reasons of supporting Docker core.

Thanks for your answer!

P.S. i already made this Dockerfile:
<//hub.docker.com/superpaintman/docker>

ikamensh:
I wish there was an official image with both docker and bash. Maintaining a separate image for this seems like unneeded user pain.

tianon:
The `apk` package manager is extremely fast and efficient - if you really want to avoid maintaining an image, you could do something like: `docker run ... docker:20.10-dind sh -c 'apk add --no-cache bash && exec dockerd-entrypoint.sh'` :eyes:

PR #30, opened 2016-11-01
Added a variant of Docker-in-Docker with docker compose

LaurentGoderre:
/cc @sylus

### DIFF
--- a/1.11/dind-compose/Dockerfile
+++ b/1.11/dind-compose/Dockerfile
@@ -1,0 +1,15 @@
+FROM docker:1.11-dind
+
+RUN apk add --no-cache \
+		'py-pip==8.1.2-r0'
+
+RUN set -x \
+	&& pip install 'docker-compose'
+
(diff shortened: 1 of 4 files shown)


tianon:
Can you explain a bit what the use case for having `docker-compose` inside an image which runs the Docker daemon is?  Normally this'd be accomplished either with `docker-compose` on your host or in a second container (see also [`docker/compose`](https://hub.docker.com/r/docker/compose/)) pointing at the first.

LaurentGoderre:
Certainly and thanks for the quick reply. We need this for our ci that uses dind as the image and we want to test our docker-compose setup

tianon:
Closing this for now in favor of https://github.com/docker/compose/issues/4287 -- we can re-open/reconsider when we hear back from the Compose maintainers on whether this is something they'd like us to pursue.  Thanks! :+1:

mbana:
Hello, any plans to add `docker-compose` to the dind image?

PR #31, opened 2016-11-16
Convert over to invoking "dockerd" directly instead of relying on the "docker daemon" translation

tianon:
This is going to be especially important for 1.13+ where `docker daemon` is now officially deprecated (and in at least `-rc1`, appears to have issues: https://github.com/docker/docker/issues/28368).


tianon:
For reference, the 1.13 failure is tracked upstream at https://github.com/docker/docker/issues/28329

PR #34, opened 2016-11-30
Change `foo --help` check to `help foo` in 1.13 entrypoint

underyx:
As per https://github.com/docker/docker/pull/28773 `docker foo --help` will no longer exit with status code 1 anymore, making the entrypoint think that *everything* is a valid docker subcommand.

Note that this will only work with 1.13-rc3 upwards, which is also why tests fail.


tianon:
Nice, thanks!  I did some testing, and this method works for both 1.12 and 1.11 as well -- would you mind updating those as well for consistency, or would you rather I take care of that?

underyx:
Sure, I'll take care of it, tomorrow.

underyx:
PR updated, @tianon

PR #35, opened 2017-01-03
Update to Alpine 3.5

tianon:


tianon:
Partial revert incoming in https://github.com/docker-library/docker/pull/37 -- see https://github.com/docker-library/golang/pull/131#issuecomment-270677556 for more details.

PR #40, opened 2017-02-03
Force cgo for golang DNS resolver

dkelleher:
The default behavior of the golang net package is to use the pure go
implementation which sends DNS requests directly to the servers listed
in resolv.conf.  This breaks when running the docker client inside a
docker container that's 'linked' with other containers given that those
links utilize teh /etc/hosts to function.  This change forces golang's
netdns to use cgo which uses the system c libraries and respects
nsswitch.

Before this change any short name conflict will result in the DNS query
resolving to a remote host instead of the linked container.  For
example, docker will resolve to docker.rc.fas.harvard.edu instead of
the linked container due to the search domain in /etc/resolv.conf.

### DIFF
--- a/1.12/docker-entrypoint.sh
+++ b/1.12/docker-entrypoint.sh
@@ -12,6 +12,9 @@ if docker help "$1" > /dev/null 2>&1; then
 	set -- docker "$@"
 fi
 
+# force cgo resolver
+export GODEBUG=netdns=cgo
+
 # if we have "--link some-docker:docker" and not DOCKER_HOST, let's set DOCKER_HOST automatically
 if [ -z "$DOCKER_HOST" -a "$DOCKER_PORT_2375_TCP" ]; then
 	export DOCKER_HOST='tcp://docker:2375'
--- a/1.13/docker-entrypoint.sh
+++ b/1.13/docker-entrypoint.sh
@@ -12,6 +12,9 @@ if docker help "$1" > /dev/null 2>&1; then
 	set -- docker "$@"
 fi
 
+# force cgo resolver
+export GODEBUG=netdns=cgo
+
 # if we have "--link some-docker:docker" and not DOCKER_HOST, let's set DOCKER_HOST automatically
 if [ -z "$DOCKER_HOST" -a "$DOCKER_PORT_2375_TCP" ]; then
 	export DOCKER_HOST='tcp://docker:2375'


dkelleher:
I should also note this is documented at [golang.org](https://golang.org/pkg/net/)

tianon:
Aren't Docker's official release binaries explicitly compiled with `-tags netgo`, and thus overriding this choice directly at compilation time?

IMO it might be better to instead document adding `-e GODEBUG=netdns=cgo` and what the difference is (if we can find a simple way to verify the behavior), or add `ENV GODEBUG="netdns=cgo"` to the `Dockerfile` if it's determined that using this behavior by default does make sense (in which case, shouldn't upstream consider adding code to handle this automatically?)

dkelleher:
I didn't see where the Docker official release binaries are compiled with `-tags netgo` but I must have missed it.  The way I read the documentation this choice can always be made to runtime using the `GODEBUG=netdns=cgo` and if it is explicitly compiled that way our tests show that the `GODEBUG` variable does override this.

My argument against documentation is,  using `netgo` with a search domain can breaks the container linking which is critical to the functionality of the Docker in Docker setup.  Additionally it breaks it in a very subtle way in order to save one system thread.

I also wrested with sending this upstream but found some arguments about similar go applications which asserted that this was a problem with the application or language and could be solved other ways.  I settle on putting it here because it minimized the impact of the change but I'm totally open to sending this upstream.

jtriley:
IMHO golang should provide `cgo` as the default netdns implementation and provide a mechanism (e.g. env variable) for choosing the pure GO resolver as an optimization which it clearly is:

```
By default the pure Go resolver is used, because a blocked DNS request 
consumes only a goroutine, while a blocked C call consumes an operating 
system thread.
```

from: https://golang.org/src/net/net.go

Also the docs claim it will switch to cgo under a variety of circumstances including if `nsswitch.conf` declares something it doesn't support but this clearly has issues:

https://github.com/golang/go/issues/11450

I think this is a worthwhile change unless there's a really compelling reason not to rely on the system C libraries for dns lookups.

It's a good question whether docker upstream should also force cgo...

yosifkit:
It looks like `-tags netgo` was officially added in https://github.com/moby/moby/pull/2162 and fixed for newer go versions in https://github.com/moby/moby/pull/10087.  I am doubting they would change it since they use it to be able to create a fully static binary and we try to keep as close to upstream releases as possible.  I would suggest just using `-e GODEBUG=netdns=cgo` when using a search domain.

tianon:
See https://github.com/docker-library/docker/pull/84, which adds an appropriate `/etc/nsswitch.conf` telling Go's `netgo` resolver to check `/etc/hosts` before doing a DNS lookup, matching Debian's default configuration. :+1:

dkelleher:
Awesome!

PR #47, opened 2017-04-01
More small (/24 instead /16) CIDR subnet for dind entrypoint

iHile:
Hi, I have wasted a lot of time because 172.18.0.0/16 is too broad and my internal network in this range.

### DIFF
--- a/17.03/dind/dockerd-entrypoint.sh
+++ b/17.03/dind/dockerd-entrypoint.sh
@@ -9,6 +9,7 @@ if [ "$#" -eq 0 -o "${1#-}" != "$1" ]; then
 		--host=unix:///var/run/docker.sock \
 		--host=tcp://0.0.0.0:2375 \
 		--storage-driver=vfs \
+        --bip=172.18.0.1/24 \
 		"$@"
 fi
 


tianon:
This is something we ought to document better -- if you run the image with `--bip=...`, it should work as-is (and I'd prefer to avoid straying too far into our own custom defaults):
```console
$ docker run ... docker:dind --bip=a.b.c.d/c
```

iHile:
gitlab can't change settings for service container (((

tianon:
That's a shame, and IMO GitLab should figure out a way to fix that. :disappointed:

In the meantime, what I'd recommend is a short `Dockerfile` that builds a custom image which is then used; something like the following:
```dockerfile
FROM docker:dind
CMD ["--bip=a.b.c.d/c"]
```

iHile:
I know, but /18 is VERY BIG NETWORK 

IMHO.

tianon:
Yeah, I tend to agree, but this is Docker's default (which we do try to stick closely to), so I'd recommend taking it up with upstream if you'd like to see the default changed (or simply create your own image which modifies the configuration as I've described either via `--bip` or via `/etc/docker/daemon.json` with `"bip": "a.b.c.d/e"`).  Thanks!

PR #63, opened 2017-06-07
Add multiarch support (again) and update to 17.06.0-ce-rc2

tianon:
See also https://github.com/docker-library/buildpack-deps/pull/59 and https://github.com/docker-library/golang/pull/163.

### DIFF
--- a/.architectures-lib
+++ b/.architectures-lib
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+_awkArch() {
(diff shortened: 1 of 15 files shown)


tianon:
Unfortunately, `library/alpine` doesn't have anything except `amd64` at the moment, so the generated `Architectures` field isn't very exciting, but it will be! :sweat_smile:

yosifkit:
image

LGTM

PR #64, opened 2017-06-13
Add a little "paperclip" suggestion when attempting to run "dockerd" in the non-dind image variants

### LINKED ISSUE #61: Trying to run "dockerd" in the non-dind image is likely to fail
As reported by @stevvooe over in https://github.com/docker-library/official-images/issues/2654, it's a little bit non-obvious that the `dind` variants are the only place we have the dependencies for properly running Docker-in-Docker (and that the non-`dind` tags are simply the client).

He suggested having `docker-entrypoint.sh` (not `dockerd-entrypoint.sh`) print a warning if `$1` is `dockerd`, something like:
```
📎 Hey there!  It looks like you're trying to run a Docker daemon.
   You probably should use the "dind" image variant instead, something like:

      docker run ...
```

tianon:
Fixes #61

Here's the example output:
```console
$ docker run -it --rm docker:17.06-rc dockerd

📎 Hey there!  It looks like you're trying to run a Docker daemon.

   You probably should use the "dind" image variant instead, something like:

     docker run --privileged --name some-overlay-docker -d docker:stable-dind --storage-driver=overlay

   See https://hub.docker.com/_/docker/ for more documentation and usage examples.

WARN[0000] could not change group /var/run/docker.sock to docker: group docker not found
...
```

### DIFF
--- a/17.03-rc/docker-entrypoint.sh
+++ b/17.03-rc/docker-entrypoint.sh
@@ -17,4 +17,19 @@ if [ -z "$DOCKER_HOST" -a "$DOCKER_PORT_2375_TCP" ]; then
 	export DOCKER_HOST='tcp://docker:2375'
 fi
 
+if [ "$1" = 'dockerd' ]; then
+	cat >&2 <<-'EOW'
+
+		📎 Hey there!  It looks like you're trying to run a Docker daemon.
+
+		   You probably should use the "dind" image variant instead, something like:
+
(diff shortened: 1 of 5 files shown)


tianon:
Worth noting (IMO) is that folks who know what they're doing can trivially get around this entire thing by using `--entrypoint dockerd` instead. :+1:

PR #68, opened 2017-07-18
dind: add git to the dind image

djs55:
Docker will shell out to `git` if it receives a request like:

```
$ curl  -d "" http://localhost:2375/v1.22/build?remote=https%3A%2F%2Fgithub.com%2Fdnephin%2Fdocker-build-from-url.git
failed to init repo at /var/lib/docker/tmp/docker-build-git645459714: : exec: "git": executable file not found in $PATH
```

or via a compose file:
```
$ cat > docker-compose.yml <<EOT
version: '2'

services:
  test:
    build: "https://github.com/dnephin/docker-build-from-url.git"
EOT

$ docker-compose build
```
This patch attempts to add git to the dind image to fix these cases.

I modified the file marked `.template` but I don't know how/if the other files should be updated. I've not tested a build from this repo, but I tested a very similar patch in [linuxkit/linuxkit#2201]

(It's a shame to make the image bigger.)

## References

- original discussion over on https://github.com/linuxkit/linuxkit/pull/2201
- [builder/remotecontext/git/gitutils.go](https://github.com/moby/moby/blob/72959fc2161fdcd785ad54a38cd03049250236fb/builder/remotecontext/git/gitutils.go#L152) which shells out to `git` (thanks to @ijc for pointing me at this)

### DIFF
--- a/Dockerfile-dind.template
+++ b/Dockerfile-dind.template
@@ -5,6 +5,7 @@ RUN apk add --no-cache \
 		btrfs-progs \
 		e2fsprogs \
 		e2fsprogs-extra \
+		git \
 		iptables \
 		xfsprogs \
 		xz


tianon:
Given that the Docker daemon invoking `git` is only intended for older API versions IIRC, I'm hesitant to add this.  See https://github.com/moby/moby/pull/4364 for where this was originally moved from daemon to the client (especially for supporting using the client's SSH keys to clone+build private repositories).

We do have a `docker:git` variant which is the client with `git` installed (separate because installing `git` is a pretty hefty size difference which most uses of the client image don't require).  I'd be willing to consider a `docker:dind-git` variant, but I'm still dubious of the value, especially since it's only still supported for older API version compatibility AFAIK.

ijc:
According to https://github.com/linuxkit/linuxkit/pull/2201#issuecomment-315246363 a build via the API (i.e. by compose rather than `docker build`) can still trigger the daemon side use of git, and that is not (currently?) deprecated. Perhaps @tonistiigi or @thaJeztah can confirm?

I did some (rough) analysis of the size increase in https://github.com/linuxkit/linuxkit/pull/2201#issuecomment-315308446, @ncopa was also amenable (not of that issue, I think it must have been on slack) to moving some of the things I highlighted into other git subpackages to reduce the overhead further.

thaJeztah:
> a build via the API (i.e. by compose rather than docker build) can still trigger the daemon side use of git, and that is not (currently?) deprecated.

Correct, it's not deprecated, but perhaps we should, as having the _daemon_ do a Git clone may not be desirable, especially given that the daemon could use credentials / keys that are present on the daemon host. t.b.d.

mbana:
Yes, please add a dind and git variant.
Thank you,

tianon:
Closing given the discussion above -- it really isn't right for the Docker daemon to be invoking `git` (that's much more correct and useful client-side, which is why Docker changed that default behavior), and it's trivial for users to add `git` to their own instance of the `dind` image (even via command if they so require with something simple like `docker run ... docker:dind sh -c 'apk add --no-cache git && exec dockerd'`), so the maintenance overhead for adding a `dind-git` variant and/or the size overhead of adding `git` to the existing `dind` variant isn't well justified.

PR #69, opened 2017-07-26
Add 17.07.0-ce-rc1

tianon:
This also removes 17.03, which should receive no further updates.


tianon:
This also adds `arm64v8`, since 17.07 has static binaries! https://download.docker.com/linux/static/test/aarch64/

PR #75, opened 2017-09-09
[WIP] add windows dockerfile

friism:
1. The next Windows version has support for bind-mounting the Windows Docker engine named pipe
2. This means it's meaningful to have an image with just `docker.exe` in it
3. In this new version, `nanoserver` doesn't have PowerShell anymore, so you're fully reliant on a builder image to get set up: https://docs.microsoft.com/en-us/windows-server/get-started/nano-in-semi-annual-channel

This PR can be tested with:

1. requires late-model windows insider (I tested on 16278)
2. ` Get-Content .\17.09-rc\windows\Dockerfile | docker build -t docker -`
3. 

```
docker run  -v \\.\pipe\docker_engine:\\.\pipe\docker_engine docker version
Client:
 Version:      17.09.0-ce-rc1
 API version:  1.31
 Go version:   go1.8.3
 Git commit:   ae21824
 Built:        Wed Sep  6 22:25:14 2017
 OS/Arch:      windows/amd64

Server:
 Version:      17.09.0-ce-rc1
 API version:  1.32 (minimum version 1.24)
 Go version:   go1.8.3
 Git commit:   ae21824
 Built:        Wed Sep  6 22:35:40 2017
 OS/Arch:      windows/amd64
 Experimental: true
```

PR #85, opened 2017-10-13
Add initial .appveyor.yml

tianon:
This is the follow-up companion to https://github.com/docker-library/docker/pull/75. :+1:


tianon:
:muscle: :metal:

PR #87, opened 2017-10-27
Add zfs progs to :dind images

AaronFriel:
Like it says on the tin. Without the zfs binaries included, the `:dind` images fail when Docker uses the zfs graph driver. If there are compatibility issues with adding the zfs package to every image, please let me know.

Here is what I have to do to get GitLab CI to work with Docker and the zfs graph driver.

1. Create a custom `:dind` image:

https://gitlab.frielforreal.com/global/docker/blob/master/dind/Dockerfile:
```dockerfile
FROM docker:dind

RUN apk add --no-cache --repository http://dl-cdn.alpinelinux.org/alpine/edge/main \
	zfs

COPY dind-entrypoint.sh /usr/local/bin/

ENTRYPOINT ["dind-entrypoint.sh"]
CMD []
```

2. Modify `.gitlab-ci.yml` to use the above images instead of `docker:dind` and alias it to ` docker`:

https://gitlab.frielforreal.com/global/docker/blob/master/.gitlab-ci.yml
```yaml
image: docker:latest

services:
  - name: registry.gitlab.frielforreal.com/global/docker:dind
    alias: docker


# ...
```

### DIFF
--- a/17.06/dind/Dockerfile
+++ b/17.06/dind/Dockerfile
@@ -7,7 +7,8 @@ RUN apk add --no-cache \
 		e2fsprogs-extra \
 		iptables \
 		xfsprogs \
-		xz
+		xz \
+		zfs
 
 # TODO aufs-tools
 
--- a/17.09/dind/Dockerfile
+++ b/17.09/dind/Dockerfile
@@ -7,7 +7,8 @@ RUN apk add --no-cache \
 		e2fsprogs-extra \
 		iptables \
 		xfsprogs \
-		xz
+		xz \
+		zfs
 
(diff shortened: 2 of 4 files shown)


tianon:
Just for historical interest, this adds ~4MB to the image size.

LGTM :+1:

(Just a note, if your custom `dind-entrypoint.sh` is only to adjust `--storage-driver`, you can simply use `CMD` to add `--storage-driver` instead. :+1:)

AaronFriel:
@tianon I actually use my entrypoint to avoid ever setting `--storage-driver`, that way Docker will resolve that automatically.

PR #91, opened 2017-11-01
Exclude "zfs" tools from arches where they're unavailable

tianon:


tianon:
For posterity, here's how this error exhibits:
```console
Step 2/10 : RUN apk add --no-cache 		btrfs-progs 		e2fsprogs 		e2fsprogs-extra 		iptables 		xfsprogs 		xz 		zfs
 ---> Running in b01c60afed09
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/main/ppc64le/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.6/community/ppc64le/APKINDEX.tar.gz
ERROR: unsatisfiable constraints:
  zfs (missing):
    required by: world[zfs]
Removing intermediate container b01c60afed09
The command '/bin/sh -c apk add --no-cache 		btrfs-progs 		e2fsprogs 		e2fsprogs-extra 		iptables 		xfsprogs 		xz 		zfs' returned a non-zero code: 1
```

tianon:
Doh, should've tested this on `ppc64le`; `apk info --no-cache zfs` returns true, even though our current arch doesn't have `zfs`. :man_facepalming:

Need a better way to detect or conditionalize this.

PR #96, opened 2017-12-01
$DOCKERD_OPTIONS

multi-io:
This introduces the ability to pass arbitrary command line options to the in-container dockerd via the container's environment ($DOCKERD_OPTIONS).


yosifkit:
Duplicate of #62 (also #12 and #20).

PR #117, opened 2018-07-06
Update pid deletion code to also delete "docker-containerd.pid"

tianon:
This is the root behind the system failure which caused https://github.com/gliderlabs/docker-alpine/issues/418#issuecomment-403054625.


tianon:
This exhibits as a log message similar to:

```
time="2018-07-05T13:50:26.765931975Z" level=info msg="libcontainerd: previous instance of containerd still alive (27)" 
```

PR #134, opened 2018-11-08
Bump 18.09 to GA

yosifkit:
https://github.com/docker/docker-ce/releases/tag/v18.09.0


.travis.yml
```diff
@@ -2,7 +2,7 @@ language: bash
 services: docker
 
 env:
-  - VERSION=18.09-rc
+  - VERSION=18.09
   - VERSION=18.06
```
tianon: @thaJeztah got a crystal ball that'll tell us if 18.06 (CE) will see any more point releases? :trollface:

thaJeztah: Oh, not sure actually; I think there's some patches that are staged for 18.06, but if they're serious enough to warrant a patch release would have to be decided 😅

If there's a particular patch that you think warrants a patch release, let me know 🤗

PR #152, opened 2019-04-15
prefer posix shell built-ins over `which`

jhoblitt:
As the current docker images are based on alpine, and busybox provides a
`which` command, this is essentially a no-op.  However, I believe it is still
worthwhile as `which` is not included by many minimal images and `shellcheck`
issues a warning on the usage of `which` by default.

### DIFF
--- a/18.06/dind/dockerd-entrypoint.sh
+++ b/18.06/dind/dockerd-entrypoint.sh
@@ -13,7 +13,7 @@ fi
 
 if [ "$1" = 'dockerd' ]; then
 	# if we're running Docker, let's pipe through dind
-	set -- "$(which dind)" "$@"
+	set -- "$(command -v dind)" "$@"
 
 	# explicitly remove Docker's default PID file to ensure that it can start properly if it was stopped uncleanly (and thus didn't clean up the PID file)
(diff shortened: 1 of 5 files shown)


tianon:
`command -v` is not 100% equivalent to `which` here though.  We need to *always* get back a file path, which is why we went for `which` instead of `command -v` (the latter of which might return a shell built-in, as you've mentioned).

jhoblitt:
Would you prefer `type -P`?

tianon:
Honestly, I'd prefer to ditch the `dind` wrapper entirely (or simply hard
code the path).  We only used `which` because we were trying to be a
*little* more generic.

I think something like `if [ -f /usr/local/bin/dind ]; then set --
/usr/local/bin/dind "$@"; fi` would be plenty sufficient if we're really
needing to eke out this one fork/exec.

jhoblitt:
I agree that being deterministic would be preferred.  As I've been looking at this for only ~15mins, I'm not sure I want to try to remove dind as my first act.  Would you consider accepting a PR to fix the path?

tianon:
Sure, sounds reasonable (although please do include an `if` test for
whether the file exists).

jhoblitt:
Take 2 opened as #154.

PR #153, opened 2019-04-15
add DOCKER_HOST as a build-arg

jhoblitt:
The current default is for dockerd to be world accessible.  This is
sub-optimal for many use scenarios and can not be overridden without
editing the Dockerfile. Per `docker info`:

    WARNING: API is accessible on http://0.0.0.0:2375 without encryption.
             Access to the remote API is equivalent to root access on the host. Refer
             to the 'Docker daemon attack surface' section in the documentation for
             more information: https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface

This changset preserves the existing default build behavior while
enabling the construction of more restricted image. Eg.,

    docker build . --build-arg DOCKER_HOST=tcp://127.0.0.1:2375

### DIFF
--- a/18.06/dind/Dockerfile
+++ b/18.06/dind/Dockerfile
@@ -1,5 +1,8 @@
 FROM docker:18.06
 
+ARG DOCKER_HOST
+ENV DOCKER_HOST=$DOCKER_HOST
+
 # https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
 RUN set -eux; \
 	apk add --no-cache \
--- a/18.06/dind/dockerd-entrypoint.sh
+++ b/18.06/dind/dockerd-entrypoint.sh
@@ -1,13 +1,15 @@
 #!/bin/sh
 set -e
 
+DOCKER_HOST=${DOCKER_HOST:-tcp://0.0.0.0:2375}
+
 # no arguments passed
 # or first arg is `-f` or `--some-option`
 if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	# add our default arguments
 	set -- dockerd \
 		--host=unix:///var/run/docker.sock \
-		--host=tcp://0.0.0.0:2375 \
+		--host="${DOCKER_HOST}" \
 		"$@"
 fi
 
--- a/18.09-rc/dind/Dockerfile
+++ b/18.09-rc/dind/Dockerfile
@@ -1,5 +1,8 @@
 FROM docker:18.09-rc
 
+ARG DOCKER_HOST
+ENV DOCKER_HOST=$DOCKER_HOST
+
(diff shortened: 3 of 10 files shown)


yosifkit:
Duplicate of #62 (also #12 and #20 and #96).

As for not being able to turn off the default, that can be done by skipping the entrypoint script:

```console
$ docker run -d --entrypoint dockerd docker:dind '--host=[Insert value here]'
```

jhoblitt:
@yosifkit Following that logic, `dockerd` could be the default entrypoint and `dind` / `dockerd-entrypoint.sh` could be removed, right?

tianon:
Essentially yes, except for the `--host` default that allows it to listen via TCP by default.

There used to be a time where `/usr/local/bin/dind` was necessary for Docker-in-Docker to work, but it's *long* behind us now (somewhere back in the 1.x releases, IIRC).

yosifkit:
> `dockerd` could be the default entrypoint and `dind` / `dockerd-entrypoint.sh` could be removed, right?

Without possibly breaking users? No.

We provide a minimal entrypoint for the most common use case of accessing dind from both the unix socket within the container (`docker exec`) or another container, with the minor helper to delete the pid on restarts of the container.

`tcp://127.0.0.1:2375` is not needed to "secure" the running container. Unless you [publish the port](https://docs.docker.com/engine/reference/commandline/run/#publish-or-expose-port--p---expose) to the Docker host, it will not be accessible to anything outside the [docker network](https://docs.docker.com/network/) in which it resides.

jhoblitt:
@yosifkit I discovered the port was world accessible as it was accessible across the network from other k8s pods (eks default cni) AND I need to be able to deploy on a shared-use k8s cluster that doesn't support network policies.  Contrary to your assertion, the image is very much unsafe out of the box in a generic kubernetes deployment scenario as a side-car container.

As an end user, I generally in the position where I need to assume that the various wrapper scripts in an image were put there for a reason by someone with more domain specific knowledge than myself.  If that isn't the case, a README or code comment pointer would be extremely helpful (and likely cut down on dups).  I certainly had no idea that `dockerd` was now able to be a direct entrypoint.

tianon:
Under https://hub.docker.com/_/docker#start-a-daemon-instance:

"By default, the dind variants of this image add --host=tcp://0.0.0.0:2375
(on top of the explicit default of --host=unix:///var/run/docker.sock) in
order to allow external containers to access dockerd appropriately (as the
following examples illustrate)."

jhoblitt:
@tianon I'm certainly willing to believe that I'm an illiterate asshat but I don't see any language that suggests `dockerd` could/should be used as an entrypoint.  The string "entrypoint" isn't present at all in the DESCRIPTION.

jhoblitt:
I think there are some other issues with the DESCRIPTION too. For example, it states that `vfs` is the default storage driver where as I seem to be getting `overlay2` as the default.

PR #154, opened 2019-04-15
use absolute path instead of dynamic lookup

jhoblitt:
This is both more deterministic and makes `shellcheck` happy.

### DIFF
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -12,8 +12,10 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 fi
 
 if [ "$1" = 'dockerd' ]; then
-	# if we're running Docker, let's pipe through dind
-	set -- "$(which dind)" "$@"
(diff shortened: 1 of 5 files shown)


dockerd-entrypoint.sh
```diff
@@ -13,7 +15,7 @@ fi
 
 if [ "$1" = 'dockerd' ]; then
 	# if we're running Docker, let's pipe through dind
-	set -- "$(which dind)" "$@"
+	[ -x "$DIND" ] && set -- "$DIND" "$@"
```
tianon: I don't see a strong reason for an extra variable here, so I'd rather see something even more straightforward like this:

```suggestion
	if [ -x /usr/local/bin/dind ]; then
		# if we have the (mostly defunct now) Docker-in-Docker wrapper script, use it
		set -- /usr/local/bin/dind "$@"
	fi
```

jhoblitt: My preference is for DRY as that tends to make future refactoring easier but as you like.

tianon:
Thanks! :+1:

PR #160, opened 2019-06-20
Remove 18.09-rc

J0WI:


tianon:
See https://github.com/docker-library/docker/commit/30147db for a good example of why this directory still exists (and `generate-stackbrew-library.sh` here is smart enough to exclude it from the output unless it's actually got something meaningful in it).

tianon:
(Once 18.09 is no longer supported, I'm fully +1 on removing this. :+1:)

J0WI:
Oh, I didn't notice that this is still in use for minor releases.

PR #165, opened 2019-07-11
Add dind-rootless

AkihiroSuda:
Usage:
```
$ docker build -t dind-rootless .
$ docker run -d --name dind-rootless --privileged dind-rootless
$ docker exec dind-rootless docker info
```

* The daemon runs in an unprivileged user with ID 1000
* `--privileged` is still required due to seccomp, apparmor, procfs, and sysfs stuff
- [ ] `-H tcp://....` will be supported soon: https://github.com/moby/moby/pull/39493
- [ ] known not to work on Fedora currently due to a containerd issue https://github.com/docker-library/docker/pull/165#issuecomment-511717143 (PR: https://github.com/containerd/containerd/pull/3419)

Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>

PR #166, opened 2019-07-17
Add automatic TLS generation (and enable it by default in 19.03+)

### LINKED ISSUE #164: Generate TLS on boot
The dind daemon entrypoint script should support generating TLS keys for the daemon and the clients, and set appropriate dockerd flags when an environment variable like `DIND_TLS_SAN` is specified.

e.g.
```console
$ docker run -d --privileged --name some-docker -e DIND_TLS_SAN=DNS:docker,DNS:localhost docker:dind
$ docker cp some-docker:/certs/for-client ./certs-for-client
$ docker run --rm --link some-docker:docker -v $(pwd)/certs-for-client:/root/.docker -e DOCKER_HOST=tcp://docker:2376 docker info
```

tianon:
Closes #164
Refs https://github.com/docker-library/docs/pull/1525

This adds a `DOCKER_TLS_CERTDIR` environment variable that, when present, will auto-enable TLS on `dockerd` by default (and will set the appropriate client flags if the necessary certificates exist).

It will attempt to generate a suitable `subjectAltName` extension value based on all available container IP addresses and hostnames, but the default generation can be extended via the `DOCKER_TLS_SAN` environment variable (in the standard OpenSSL format, ala `IP:n.n.n.n,DNS:foobar,...`).

For users of 18.09 who wish to enable this behavior, simply set `DOCKER_TLS_CERTDIR` to a path within the container into which you want certificates generated (and share at least the `client` subdirectory of that path with your client containers). The default value in 19.03+ is `/certs` (so to mimic that, something like `-e DOCKER_TLS_CERTDIR=/certs` would be sufficient/appropriate).

For users of 19.03+ who wish to *disable* this behavior (not recommended), simply set `DOCKER_TLS_CERTDIR` to the empty string (`-e DOCKER_TLS_CERTDIR=`).

### DIFF
--- a/18.09-rc/dind/dockerd-entrypoint.sh
+++ b/18.09-rc/dind/dockerd-entrypoint.sh
@@ -1,14 +1,119 @@
 #!/bin/sh
-set -e
+set -eu
+
+_tls_ensure_private() {
+	local f="$1"; shift
+	[ -s "$f" ] || openssl genrsa -out "$f" 4196
+}
+_tls_san() {
+	{
+		ip -oneline address | awk '{ gsub(/\/.+$/, "", $4); print "IP:" $4 }'
+		{
+			cat /etc/hostname
+			echo 'docker'
+			echo 'localhost'
+			hostname -f
+			hostname -s
+		} | sed 's/^/DNS:/'
+		[ -z "${DOCKER_TLS_SAN:-}" ] || echo "$DOCKER_TLS_SAN"
+	} | sort -u | xargs printf '%s,' | sed "s/,\$//"
+}
+_tls_generate_certs() {
+	local dir="$1"; shift
+
+	# if ca/key.pem || !ca/cert.pem, generate CA public if necessary
+	# if ca/key.pem, generate server public
+	# if ca/key.pem, generate client public
+	# (regenerating public certs every startup to account for SAN/IP changes and/or expiration)
+
+	# https://github.com/FiloSottile/mkcert/issues/174
+	local certValidDays='825'
+
+	if [ -s "$dir/ca/key.pem" ] || [ ! -s "$dir/ca/cert.pem" ]; then
+		# if we either have a CA private key or do *not* have a CA public key, then we should create/manage the CA
+		mkdir -p "$dir/ca"
+		_tls_ensure_private "$dir/ca/key.pem"
+		openssl req -new -key "$dir/ca/key.pem" \
+			-out "$dir/ca/cert.pem" \
+			-subj '/CN=docker:dind CA' -x509 -days "$certValidDays"
+	fi
+
+	if [ -s "$dir/ca/key.pem" ]; then
+		# if we have a CA private key, we should create/manage a server key
+		mkdir -p "$dir/server"
+		_tls_ensure_private "$dir/server/key.pem"
+		openssl req -new -key "$dir/server/key.pem" \
+			-out "$dir/server/csr.pem" \
+			-subj '/CN=docker:dind server'
+		cat > "$dir/server/openssl.cnf" <<-EOF
+			[ x509_exts ]
+			subjectAltName = $(_tls_san)
+		EOF
+		openssl x509 -req \
+				-in "$dir/server/csr.pem" \
+				-CA "$dir/ca/cert.pem" \
+				-CAkey "$dir/ca/key.pem" \
+				-CAcreateserial \
+				-out "$dir/server/cert.pem" \
+				-days "$certValidDays" \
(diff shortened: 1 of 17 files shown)


tianon:
Test updated in https://github.com/docker-library/official-images/commit/7e44f391f28ecba7821b5ded0c197eb99612f6d3 :+1:

tianon:
(We should probably also add a new `docker-dind-insecure` test that just does the basics to verify `DOCKER_TLS_CERTDIR=` works to disable this.)

tianon:
(dat +529 −40 tho :sob:)

18.09-rc/dind/dockerd-entrypoint.sh
```diff
@@ -1,14 +1,116 @@
...
+	# if ca/key.pem, generate client public
+	# (regenerating public certs every startup to account for SAN/IP changes and/or expiration)
+
+	if [ -s "$dir/ca/key.pem" ] || [ ! -s "$dir/ca/cert.pem" ]; then
+		# if we either have a CA private key or do *not* have a CA public key, then we should create/manage the CA
+		mkdir -p "$dir/ca"
+		_tls_ensure_private "$dir/ca/key.pem"
+		openssl req -new -key "$dir/ca/key.pem" \
+			-out "$dir/ca/cert.pem" \
+			-subj '/CN=docker:dind CA' -x509 -days "$(( 365 * 30 ))"
```
AkihiroSuda: should be 825 days? https://github.com/FiloSottile/mkcert/issues/174

18.09-rc/dind/dockerd-entrypoint.sh
```diff
@@ -1,14 +1,116 @@
...
+			[ x509_exts ]
+			subjectAltName = $(_tls_san)
+		EOF
+		openssl x509 -req \
+				-in "$dir/server/csr.pem" \
+				-CA "$dir/ca/cert.pem" \
+				-CAkey "$dir/ca/key.pem" \
+				-CAcreateserial \
+				-out "$dir/server/cert.pem" \
+				-days "$(( 365 * 30 ))" \
```
AkihiroSuda: should be 825 days? FiloSottile/mkcert#174

19.03-rc/Dockerfile
```diff
@@ -49,5 +49,8 @@ RUN set -eux; \
 COPY modprobe.sh /usr/local/bin/modprobe
 COPY docker-entrypoint.sh /usr/local/bin/
 
+# DOCKER_TLS_CERTDIR is where certificates for TLS will be automatically pulled from (and generated into for dind)
```
AkihiroSuda: this comment seems confusing, because /certs needs to be mounted manually into client containers

18.09-rc/dind/dockerd-entrypoint.sh
```diff
@@ -1,14 +1,116 @@
...
+			cat /etc/hostname
+			echo 'docker'
+			echo 'localhost'
+			hostname -f
+			hostname -s
+		} | sed 's/^/DNS:/'
+		[ -z "${DOCKER_TLS_SAN:-}" ] || echo "$DOCKER_TLS_SAN"
+	} | sort -u | xargs printf '%s,' | sed "s/,\$//"
+}
+_tls_generate_certs() {
```
AkihiroSuda: any chance to split this to a separate script so that it can be called from rootless entrypoint script? https://github.com/docker-library/docker/pull/165

18.09-rc/docker-entrypoint.sh
```diff
@@ -12,9 +12,24 @@ if docker help "$1" > /dev/null 2>&1; then
 	set -- docker "$@"
 fi
 
-# if we have "--link some-docker:docker" and not DOCKER_HOST, let's set DOCKER_HOST automatically
-if [ -z "$DOCKER_HOST" -a "$DOCKER_PORT_2375_TCP" ]; then
-	export DOCKER_HOST='tcp://docker:2375'
+_should_tls() {
+	[ -n "${DOCKER_TLS_CERTDIR:-}" ] \
+	&& [ -s "$DOCKER_TLS_CERTDIR/client/ca.pem" ] \
```
AkihiroSuda: If the client is expected to mount the parent directory of `client` (and `server`), the server key is likely to be leaked into the client

AkihiroSuda:
opened https://github.com/docker-library/docker/pull/168 for splitting tls script

tianon:
Documentation PR is now open at https://github.com/docker-library/docs/pull/1538 :+1:

PR #168, opened 2019-07-18
split tls-generate-certs.sh from dockerd-entrypoint.sh

AkihiroSuda:

so that it can be called from dind-rootless (#165) entrypoint script

### DIFF
--- a/18.09-rc/dind/Dockerfile
+++ b/18.09-rc/dind/Dockerfile
@@ -36,6 +36,7 @@ RUN set -eux; \
 	wget -O /usr/local/bin/dind "https://raw.githubusercontent.com/docker/docker/${DIND_COMMIT}/hack/dind"; \
 	chmod +x /usr/local/bin/dind
 
+COPY tls-generate-certs.sh /usr/local/bin/
 COPY dockerd-entrypoint.sh /usr/local/bin/
 
 VOLUME /var/lib/docker
--- a/18.09-rc/dind/dockerd-entrypoint.sh
+++ b/18.09-rc/dind/dockerd-entrypoint.sh
@@ -1,100 +1,12 @@
 #!/bin/sh
 set -eu
 
-_tls_ensure_private() {
-	local f="$1"; shift
-	[ -s "$f" ] || openssl genrsa -out "$f" 4196
-}
-_tls_san() {
-	{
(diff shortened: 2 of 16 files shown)


AkihiroSuda:
An alternative way is to just extend the current dockerd-entrypoint.sh to support rootless

tianon:
I think I'd like to understand more about the goal of running rootless `dockerd` inside a container before committing to any adjustment here one way or another -- given that it still requires `--privileged`, what's gained by doing so?

AkihiroSuda:
Rootless protects the host from potential vulns of Docker, containerd, and runc

tianon:
But if it's running inside a `--privileged` container, it still has more elevated privileges than non-root on the host, right?

AkihiroSuda:
No, it is just same as non-root on the host.

`--privileged` is required for disabling seccomp/apparmor so as to allow `unshare(2)` and other container-related syscalls, and for unmasking /proc/* and /sys/* so that nested containers can mount /proc with a new pid namespace and /sys with a new net namespace.

In other words, `--privileged` is required only for elevating non-root in a container to non-root on the host.

AkihiroSuda:
Or maybe it is less privileged than non-root-on-the-host, because  it has fewer suid binaries (only `/usr/bin/newuidmap` and `/usr/bin/newgidmap`)

AkihiroSuda:
Also, as a side effect, splitting the script will be helpful for testing issues like https://github.com/docker-library/docker/issues/171

AkihiroSuda:
rebased

yosifkit:
I'd rather not have a sprawl of many scripts and instead have one script that can do everything required for dind-related startup (root or not).

AkihiroSuda:
The standard `dind` tag image should contain all rootless dependencies (vpnkit and rootlesskit) as well?

AkihiroSuda:
@tonistiigi @tiborvass Do you think the rootless image should be separated from `dind`, or it should be just `dind` with `--user 1000`?

tianon:
It's going to be separate for (the `rootlesskit` and `vpnkit` binaries are
not exactly trivial in size).

I've got an alternate proposal to yours worked up, I just haven't had time
to actually open it as a PR for you to review yet. Hoping to get that open
ASAP (either tonight or tomorrow).

AkihiroSuda:
closing in favor of #174

Issue #170, opened 2019-07-23
latest 'dind' tag (19.03) gives error on Gitlab CI "failed to dial gRPC: cannot connect to the Docker daemon. Is 'docker daemon' running on this host?"


daenny:
We are running a gitlab server and several gitlab-ci-runners. Today we woke up to several failed builds. 
We did several tests and we found out that the most likely culprits are the newest tags of the `docker:dind` and `docker:git` images. 
We tested with `docker:18-dind`  and `docker:18-git` and the errors does not occur anymore. 

The error is given below:
> time="2019-07-23T06:52:31Z" level=error msg="failed to dial gRPC: cannot connect to the Docker daemon. Is 'docker daemon' running on this host?: dial tcp 172.17.0.3:2375: connect: connection refused"

The gitlab-runners are running in privileged mode. 

EDIT: This is not a bug or unresolved issue:
see: https://github.com/docker-library/docker/issues/170#issuecomment-514366149

RuriRyan:
I guess we have the same issue.

Errors by the actual docker:stable-dind iamge:

```bash
mount: permission denied (are you root?)
Could not mount /sys/kernel/security.
AppArmor detection and --privileged mode might break.
mount: permission denied (are you root?)
```

jubel-han:
We are facing the same issue, It seems that docker daemon is listening on `2376` instead of `2375`

```
INFO[2019-07-23T07:30:10.485681000Z] Docker daemon                                 commit=aeac9490dc graphdriver(s)=overlay2 version=19.03.0
INFO[2019-07-23T07:30:10.486308400Z] Daemon has completed initialization
INFO[2019-07-23T07:30:10.513031700Z] API listen on [::]:2376
INFO[2019-07-23T07:30:10.513069100Z] API listen on /var/run/docker.sock
```

But, after updated the `DOCKER_HOST` to `tcp://localhost:2376`, another error bob up.

```
/ # docker -v
Docker version 19.03.0, build aeac9490dc
/ # docker -D info
Client:
 Debug Mode: true

Server:
ERROR: Error response from daemon: Client sent an HTTP request to an HTTPS server.
errors pretty printing info
```

jubel-han:
Just found the workaround, exposing the environment variable `DOCKER_TLS_CERTDIR=''` would solve the problem. ref: https://github.com/docker-library/docker/pull/166

frbl:
confirming, the workaround suggested by jubel-han fixes it (tested on Gitlab CI)

[...]

PR #173, opened 2019-07-23
Remove EXPOSE completely

tianon:
Hopefully this is temporary -- apparently folks are using `EXPOSE` as a rough guide for doing healthchecks? (https://gitlab.com/gitlab-org/gitlab-runner/issues/4501#note_194904002)

### DIFF
--- a/18.09-rc/dind/Dockerfile
+++ b/18.09-rc/dind/Dockerfile
@@ -39,7 +39,10 @@ RUN set -eux; \
 COPY dockerd-entrypoint.sh /usr/local/bin/
 
 VOLUME /var/lib/docker
-EXPOSE 2375 2376
+
+# TODO figure out why folks are using EXPOSE for health checks (https://gitlab.com/gitlab-org/gitlab-runner/issues/4501#note_194904002)
+#EXPOSE 2375 2376
+# (just because a port is listed under EXPOSE does not mean it will actually be listened on)
 
 ENTRYPOINT ["dockerd-entrypoint.sh"]
 CMD []
--- a/18.09/dind/Dockerfile
+++ b/18.09/dind/Dockerfile
@@ -39,7 +39,10 @@ RUN set -eux; \
 COPY dockerd-entrypoint.sh /usr/local/bin/
 
 VOLUME /var/lib/docker
-EXPOSE 2375 2376
+
+# TODO figure out why folks are using EXPOSE for health checks (https://gitlab.com/gitlab-org/gitlab-runner/issues/4501#note_194904002)
+#EXPOSE 2375 2376
+# (just because a port is listed under EXPOSE does not mean it will actually be listened on)
 
 ENTRYPOINT ["dockerd-entrypoint.sh"]
 CMD []
--- a/19.03-rc/dind/Dockerfile
+++ b/19.03-rc/dind/Dockerfile
@@ -39,7 +39,10 @@ RUN set -eux; \
 COPY dockerd-entrypoint.sh /usr/local/bin/
 
 VOLUME /var/lib/docker
-EXPOSE 2375 2376
+
+# TODO figure out why folks are using EXPOSE for health checks (https://gitlab.com/gitlab-org/gitlab-runner/issues/4501#note_194904002)
+#EXPOSE 2375 2376
+# (just because a port is listed under EXPOSE does not mean it will actually be listened on)
 
 ENTRYPOINT ["dockerd-entrypoint.sh"]
 CMD []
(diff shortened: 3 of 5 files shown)


tabakhase:
I think this would create more fallout than even #170 did...  

force-disable the gitlab team could do across the whole fleet last night, so most people never saw it https://gitlab.com/gitlab-org/gitlab-runner/issues/4501#note_194648520  

for "ci failures cause buildscript starts before service is ready" there is nothing todo globally.

And given that this image does NOT have a `HEALTHCHECK`, whats the alternative?

exposing "both ports when one will never come up" seems kinda wrong indeed, but so does having no expose...
((and I dont think gitlab is alone here, `docker run -P`, portainer & co...))

tianon:
There's quite a difference between `docker run -P` (which simply maps the `EXPOSE` ports to external ports so they become accessible if they're actually used) and assuming the ports will be listened on.

In most images/software, the port that's listened on is configuration, so it depends on more factors just as it is here.  With our stock configuration, we'll listen on 2376, but it's one trivial configuration later and it's 2375, or even another and it's not listening externally at all, so `EXPOSE` cannot possibly be more than a guideline for what ports *might* be useful.

Regarding `HEALTHCHECK`, it has similar issues -- if we did add a healthcheck by default, which port would it check?  It's not really reasonable for us to provide a generic healthcheck this (https://github.com/docker-library/faq#healthcheck).

tabakhase:
> and assuming the ports will be listened on.  

sure sure ;-) - thats more regarding `having no expose...`


> `HEALTHCHECK`, it has similar issues

for 18.x+ something like attempting `docker system dial-stdio` might be acceptable?  
but I see your issue and the generally missing "ready" state in the docker-system is a well known...

----

to add to this from IRC, "by the book" i think there should also be a `VOLUME /certs`,
I agree with your "why not", but then it also gets as far as almost granting 2 images...  
- `*-dind-tls` with volume and 2376 
- `*-dind-tcp` with 2375

tmaczukin:
@tianon Yes, we use port information for doing health checks because of the nature of GitLab CI services. We define services as a _network accessible service that can be use by your job_.

And to make users live easier, we check if the service is listening before we start a script that is using it. In case if it's not (or we were unable to check this), we print the warning. So in case of CI job failure, users have an information about the initial service state and can use this information while deciding what caused the job to fail.

For this however we need to somehow decide how to access the service. In the past, when we've introduced the services, there was no `HEALTHCHECK` mechanism implemented in Docker. The configuration of services in `.gitlab-ci.yml` file was also very simple - you could only specify which image to use. Using information about exposed ports was the only way to handle this. And to be precise, GitLab Runner's health check isn't relaying directly on the `EXPOSE`. It's relaying on the `..._TCP_ADDR` variables that are added to linked containers. But if the container (in this - the DinD service) doesn't specify port mapping by itself, the `EXPOSE` ports are mapped by default.

Nowadays we have `HEALTHCHECK`, which could be definitely used - in that case we'd depend fully on the image author and need to believe, that the implemented health check is really checking all that is required. We have also an extended syntax for services configuration, so we could allow users to explicitly define which ports should be used for the health check implemented in the Runner.

Usage of both of these features is already something that we're considering, please check https://gitlab.com/gitlab-org/gitlab-runner/issues/3984.

I hope that the above gives a little light on why we've decided to use this information and answers your note from this PR:

```
# TODO figure out why folks are using EXPOSE for health checks (https://gitlab.com/gitlab-org/gitlab-runner/issues/4501#note_194904002)
```

If I could ask, please postpone a while with removing the EXPOSE. It doesn't change much for you, but in case of GitLab CI it will be a huge change. Now, the service startup failure warning is presented randomly - depending on which port the user is using and which one was the first found by the health check mechanism. If you will remove the EXPOSEs, it will fail the health check constantly.

tianon:
I was looking at removing `EXPOSE` to try and help the GitLab case and we held off when it looked like this wouldn't fix anything, so I'm going to close this instead. 😅👍

tianon:
In the current case, is the code in runner looking for *both* ports to be open, or just one? Is it choosing one randomly, deterministically, or trying all until one is listening?

tmaczukin:
@tianon Runner is scaning environment variables, looking for the ones that have `_TCP_ADDR` and `_TCP_PORT` suffixes. The first one found is chosen. And since there is no guarantee on variables ordering and if no explicit port mapping is done **all** ports from `EXPOSE` are exposed, the finding is done totally randomly.

It's a very naive logic, but in most cases (where service images don't expose multiple ports) it just works and was the simplest thing that we could do in the past.

This case showed how risky assumption it is and I hope we'll find some time soon to work on an extended support for the health checks.

Also thank you very much for your activity in helping us resolve problems between GitLab CI and DinD image! :)

tianon:
So perhaps a smaller (intermediate) solution could be to scan all available ports instead of just the first discovered and wait for any of them to open?

It's my pleasure, you folks are great to work with and I enjoy GitLab's product (even spending most of my days using your primary competitor :sweat_smile:).

PR #174, opened 2019-07-30
Add experimental "rootless" dind variant

tianon:
This is an alternative proposal to https://github.com/docker-library/docker/pull/165 + https://github.com/docker-library/docker/pull/168.

This should have roughly the same behavior as the `dockerd-rootless.sh` script but without the extra depdendency (and frankly I'd love to pull some official `vpnkit` release binaries too instead of pulling the extras bundle, but I couldn't find any official binary releases nor did it look trivial to build from source).  This allows us to iterate on this in between Docker releases as well, if necessary (like pulling in a newer `rootlesskit` release, for example).

I was toying with automatically adding `--experimental` as well, but wanted to enforce that being opt-in so that users don't switch without realizing what they're doing (since it does require *some* changes in usage).

Generally everything should "just work" as automatically as possible, but with the appropriate escape hatches for users who need/want to control the underlying knobs more directly.

cc @AkihiroSuda

PR #177, opened 2019-08-02
Rename variable used for defining used docker.sock file

### LINKED ISSUE #175: docker:dind tries to listen to two overlapping ports
It seems that since 7db7eccb7b7507b1a6ba3ef81c6f37b16149cc55 the `docker:dind` container listens on `$DOCKER_HOST` instead of `unix:///var/run/docker.sock` besides listening on `0.0.0.0` port `2375` or `2376`.
However, for GitLab CI to work (in Kubernetes) we have to set the environment variable `DOCKER_HOST` to `tcp://localhost:2375`. This results in `docker:dind` trying to bind to 2375 twice (if `DOCKER_TLS_CERTDIR=""` which is needed for now to get stuff working until Gitlab offers an option to share the certdir), which gives a `bind: address already in use` error.

For now I have fixed this by setting `DOCKER_HOST` to `tcp://localhost:2378`, letting Docker use a port that is not in use yet, but it seems to me that this is not an ideal solution.

Note that it not only broke for `docker:dind` but also for other versions like `docker:18-dind`.

tmaczukin:
https://github.com/docker-library/docker/pull/174 added an experimental "rootless" variant of the DinD service.

While the change itself is simple and should not affect current usages, a `DOCKER_HOST` variable was used to define which docker.sock path should be used depending on the process owner's ID.

Chosing DOCKER_HOST variable for this is unfortunately not the best option:

1. It's not the best semantic choice - the code that uses it defines a socket file, not the host. DOCKER_SOCKET just seems to fit better the purpose.

1. DOCKER_HOST is already a variable, that is used by users to define where the Docker daemon is listening. In case when it's added to the DinD container (which is a common situation for example for GitLab CI jobs that are using DinD as a service), it finally ends with assigning two times the same port, while once it uses 0.0.0.0 address, and once some unresolvable domain name (depending on what user defined; most probably `docker`).

This commit proposes a change of the variable name to DOCKER_SOCKET, which will better match it purpose and additionally it will stop breaking configurations of many of docker:dind image users.

Fixes #175

### DIFF
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -92,16 +92,16 @@ _tls_generate_certs() {
 # no arguments passed
 # or first arg is `-f` or `--some-option`
 if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
-	# set DOCKER_HOST to the default "--host" value (for both standard or rootless)
+	# set DOCKER_SOCKET to the default "--host" value (for both standard or rootless)
 	uid="$(id -u)"
 	if [ "$uid" = '0' ]; then
-		: "${DOCKER_HOST:=unix:///var/run/docker.sock}"
+		: "${DOCKER_SOCKET:=unix:///var/run/docker.sock}"
 	else
 		# if we're not root, we must be trying to run rootless
 		: "${XDG_RUNTIME_DIR:=/run/user/$uid}"
-		: "${DOCKER_HOST:=unix://$XDG_RUNTIME_DIR/docker.sock}"
+		: "${DOCKER_SOCKET:=unix://$XDG_RUNTIME_DIR/docker.sock}"
 	fi
-	export DOCKER_HOST
+	export DOCKER_SOCKET
 
 	# add our default arguments
 	if [ -n "${DOCKER_TLS_CERTDIR:-}" ] \
@@ -112,7 +112,7 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	; then
 		# generate certs and use TLS if requested/possible (default in 19.03+)
 		set -- dockerd \
-			--host="$DOCKER_HOST" \
+			--host="$DOCKER_SOCKET" \
 			--host=tcp://0.0.0.0:2376 \
 			--tlsverify \
 			--tlscacert "$DOCKER_TLS_CERTDIR/server/ca.pem" \
@@ -123,7 +123,7 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	else
 		# TLS disabled (-e DOCKER_TLS_CERTDIR='') or missing certs
 		set -- dockerd \
-			--host="$DOCKER_HOST" \
+			--host="$DOCKER_SOCKET" \
 			--host=tcp://0.0.0.0:2375 \
 			"$@"
 		DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS="${DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS:-} -p 0.0.0.0:2375:2375/tcp"
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -92,16 +92,20 @@ _tls_generate_certs() {
 # no arguments passed
 # or first arg is `-f` or `--some-option`
 if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
-	# set DOCKER_SOCKET to the default "--host" value (for both standard or rootless)
+	# set "dockerSocket" to the default "--host" *unix socket* value (for both standard or rootless)
 	uid="$(id -u)"
 	if [ "$uid" = '0' ]; then
-		: "${DOCKER_SOCKET:=unix:///var/run/docker.sock}"
+		dockerSocket='unix:///var/run/docker.sock'
 	else
 		# if we're not root, we must be trying to run rootless
 		: "${XDG_RUNTIME_DIR:=/run/user/$uid}"
-		: "${DOCKER_SOCKET:=unix://$XDG_RUNTIME_DIR/docker.sock}"
+		dockerSocket="unix://$XDG_RUNTIME_DIR/docker.sock"
 	fi
-	export DOCKER_SOCKET
+	case "${DOCKER_HOST:-}" in
+		unix://*)
+			dockerSocket="$DOCKER_HOST"
+			;;
+	esac
 
 	# add our default arguments
 	if [ -n "${DOCKER_TLS_CERTDIR:-}" ] \
@@ -112,7 +116,7 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	; then
 		# generate certs and use TLS if requested/possible (default in 19.03+)
 		set -- dockerd \
-			--host="$DOCKER_SOCKET" \
+			--host="$dockerSocket" \
 			--host=tcp://0.0.0.0:2376 \
 			--tlsverify \
 			--tlscacert "$DOCKER_TLS_CERTDIR/server/ca.pem" \
@@ -123,7 +127,7 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	else
 		# TLS disabled (-e DOCKER_TLS_CERTDIR='') or missing certs
 		set -- dockerd \
-			--host="$DOCKER_SOCKET" \
+			--host="$dockerSocket" \
 			--host=tcp://0.0.0.0:2375 \
 			"$@"
 		DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS="${DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS:-} -p 0.0.0.0:2375:2375/tcp"
--- a/18.09-rc/dind/dockerd-entrypoint.sh
+++ b/18.09-rc/dind/dockerd-entrypoint.sh
@@ -92,16 +92,16 @@ _tls_generate_certs() {
 # no arguments passed
 # or first arg is `-f` or `--some-option`
 if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
-	# set DOCKER_HOST to the default "--host" value (for both standard or rootless)
+	# set DOCKER_SOCKET to the default "--host" value (for both standard or rootless)
 	uid="$(id -u)"
 	if [ "$uid" = '0' ]; then
-		: "${DOCKER_HOST:=unix:///var/run/docker.sock}"
+		: "${DOCKER_SOCKET:=unix:///var/run/docker.sock}"
 	else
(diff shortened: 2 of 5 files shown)


tmaczukin:
@tianon Could I get your attention on this PR?

vasilerobert85:
Hi Tomasz,

I have commented out DOCKER_HOST: tcp://docker:2375 and set docker:18.09.7-dind and it seems to work.

Did not risk going to  `19.03.0-dind`

```variables:
  ECR_SCREENING: $ECR_SCREENING
  DOCKER_IMAGE_TAG: $CI_REGISTRY_IMAGE:$CI_COMMIT_SHA
  #DOCKER_HOST: tcp://docker:2375
  DOCKER_DRIVER: overlay2
  DOCKER_TLS_CERTDIR: ""

services:
  - docker:18.09.7-dind

stages:
  - build
  - test
  - push
  - deploy

build:
  image: docker:18.09.7-dind
  stage: build
  only:
    - merge_requests
    - master
  script:
    - docker login -u gitlab-ci-token -p $CI_JOB_TOKEN $CI_REGISTRY
    - docker build -t $DOCKER_IMAGE_TAG .
    - docker push $DOCKER_IMAGE_TAG```

**Comment by vasilerobert85** [comment]
tested with 19.03.0-dind as well and --> is `working`

**Comment by tmaczukin** [comment]
Also cc @AkihiroSuda and @yosifkit who were involved in reviewing and merging #174 :)

**Comment by tianon** [comment]
Ouch, I ran into a related problem last night just before going to bed but hadn't had a chance to dig into it yet. :disappointed:

I agree with changing this variable name, but I think in doing so we should ditch the `:-` and the `export` (making this a local script variable instead of an environment variable, since `DOCKER_SOCKET` doesn't have any meaning outside of here and isn't something we want to create), and we should also check whether `DOCKER_HOST` is set to something that starts with `unix://` and use that value if so (which was the original intention here, since Docker picks up that value, but I obviously forgot the other half :man_facepalming:).

There's also another path where this might get set strangely: `docker run ... docker:dind dockerd-entrypoint.sh ...` (which has been supported previously -- the updated code turns this into `dockerd-entrypoint.sh docker-entrypoint.sh dockerd-entrypoint.sh` which sets `DOCKER_HOST` and causes the same havoc which can be healed in the same way).

Here's my proposed adjustment to this:

```diff
diff --git a/dockerd-entrypoint.sh b/dockerd-entrypoint.sh
index 47eee55..cdbea49 100755
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -92,16 +92,16 @@ _tls_generate_certs() {
 # no arguments passed
 # or first arg is `-f` or `--some-option`
 if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
-	# set DOCKER_HOST to the default "--host" value (for both standard or rootless)
+	# set "dockerSocket" to the default "--host" *unix socket* value (for both standard or rootless)
 	uid="$(id -u)"
 	if [ "$uid" = '0' ]; then
-		: "${DOCKER_HOST:=unix:///var/run/docker.sock}"
+		dockerSocket='unix:///var/run/docker.sock'
 	else
 		# if we're not root, we must be trying to run rootless
 		: "${XDG_RUNTIME_DIR:=/run/user/$uid}"
-		: "${DOCKER_HOST:=unix://$XDG_RUNTIME_DIR/docker.sock}"
+		dockerSocket="unix://$XDG_RUNTIME_DIR/docker.sock"
 	fi
-	export DOCKER_HOST
+	case "${DOCKER_HOST:-}" in unix://*) dockerSocket="$DOCKER_HOST" ;; esac
 
 	# add our default arguments
 	if [ -n "${DOCKER_TLS_CERTDIR:-}" ] \
@@ -112,7 +112,7 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	; then
 		# generate certs and use TLS if requested/possible (default in 19.03+)
 		set -- dockerd \
-			--host="$DOCKER_HOST" \
+			--host="$dockerSocket" \
 			--host=tcp://0.0.0.0:2376 \
 			--tlsverify \
 			--tlscacert "$DOCKER_TLS_CERTDIR/server/ca.pem" \
@@ -123,7 +123,7 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	else
 		# TLS disabled (-e DOCKER_TLS_CERTDIR='') or missing certs
 		set -- dockerd \
-			--host="$DOCKER_HOST" \
+			--host="$dockerSocket" \
 			--host=tcp://0.0.0.0:2375 \
 			"$@"
 		DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS="${DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS:-} -p 0.0.0.0:2375:2375/tcp"
@@ -175,8 +175,9 @@ if [ "$1" = 'dockerd' ]; then
 			${DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS:-} \
 			"$@" --userland-proxy-path=rootlesskit-docker-proxy
 	fi
-else
+elif base="$(basename "$1")" && [ "$base" != 'dockerd-entrypoint.sh' ] && [ "$base" != 'docker-entrypoint.sh' ]; then
 	# if it isn't `dockerd` we're trying to run, pass it through `docker-entrypoint.sh` so it gets `DOCKER_HOST` set appropriately too
+	# ("docker run ... docker:dind dockerd-entrypoint.sh ..." is legal, and "DOCKER_HOST" affects dockerd so we don't want to set it there)
 	set -- docker-entrypoint.sh "$@"
 fi
 
```

tianon:
I guess really with the `DOCKER_HOST` fix, we can skip that second half -- it's purely an optimization at that point (and thus unrelated):

```diff
diff --git a/dockerd-entrypoint.sh b/dockerd-entrypoint.sh
index 47eee55..cdbea49 100755
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -92,16 +92,16 @@ _tls_generate_certs() {
 # no arguments passed
 # or first arg is `-f` or `--some-option`
 if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
-	# set DOCKER_HOST to the default "--host" value (for both standard or rootless)
+	# set "dockerSocket" to the default "--host" *unix socket* value (for both standard or rootless)
 	uid="$(id -u)"
 	if [ "$uid" = '0' ]; then
-		: "${DOCKER_HOST:=unix:///var/run/docker.sock}"
+		dockerSocket='unix:///var/run/docker.sock'
 	else
 		# if we're not root, we must be trying to run rootless
 		: "${XDG_RUNTIME_DIR:=/run/user/$uid}"
-		: "${DOCKER_HOST:=unix://$XDG_RUNTIME_DIR/docker.sock}"
+		dockerSocket="unix://$XDG_RUNTIME_DIR/docker.sock"
 	fi
-	export DOCKER_HOST
+	case "${DOCKER_HOST:-}" in unix://*) dockerSocket="$DOCKER_HOST" ;; esac
 
 	# add our default arguments
 	if [ -n "${DOCKER_TLS_CERTDIR:-}" ] \
@@ -112,7 +112,7 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	; then
 		# generate certs and use TLS if requested/possible (default in 19.03+)
 		set -- dockerd \
-			--host="$DOCKER_HOST" \
+			--host="$dockerSocket" \
 			--host=tcp://0.0.0.0:2376 \
 			--tlsverify \
 			--tlscacert "$DOCKER_TLS_CERTDIR/server/ca.pem" \
@@ -123,7 +123,7 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	else
 		# TLS disabled (-e DOCKER_TLS_CERTDIR='') or missing certs
 		set -- dockerd \
-			--host="$DOCKER_HOST" \
+			--host="$dockerSocket" \
 			--host=tcp://0.0.0.0:2375 \
 			"$@"
 		DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS="${DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS:-} -p 0.0.0.0:2375:2375/tcp"
```

tmaczukin:
@tianon Thanks for the update. I agree that using the local variable may be a better way. And thanks for pointing the `check whether DOCKER_HOST is set to something that starts with unix://` case, since I've totally haven thought about it.

I also think there is no need to change how the last case is handled - in context of this fix. An optimization is a separate issue (and probably it could also get some attention, but it's totally not related to #175 problem).

I've pushed an update, using your second example. What do you think now?

dockerd-entrypoint.sh
```diff
@@ -92,16 +92,20 @@ _tls_generate_certs() {
...
+                dockerSocket='unix:///var/run/docker.sock'
 	else
 		# if we're not root, we must be trying to run rootless
 		: "${XDG_RUNTIME_DIR:=/run/user/$uid}"
-		: "${DOCKER_HOST:=unix://$XDG_RUNTIME_DIR/docker.sock}"
+                dockerSocket="unix://$XDG_RUNTIME_DIR/docker.sock"
 	fi
-	export DOCKER_HOST
+        case "${DOCKER_HOST:-}"
+                in unix://*)
```
tianon: I agree with exploding this for readability, but we typically put `in` on the line with the `case` (since a given `case` will only ever have one `in` keyword).

dockerd-entrypoint.sh
```diff
@@ -92,16 +92,20 @@ _tls_generate_certs() {
 # no arguments passed
 # or first arg is `-f` or `--some-option`
 if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
-	# set DOCKER_HOST to the default "--host" value (for both standard or rootless)
+        # set "dockerSocket" to the default "--host" *unix socket* value (for both standard or rootless)
```
tianon: Can you please adjust the space indents to tabs throughout this change (matching the rest of the file)? :pray: :heart:

dockerd-entrypoint.sh
tmaczukin: You're totally right! My bad, I've used `[enter]` button in wrong place ;)

dockerd-entrypoint.sh
tmaczukin: Of course!

dockerd-entrypoint.sh
tmaczukin: Updated

dockerd-entrypoint.sh
tmaczukin: Updated

tianon:
... and Travis is already done :tada:

tmaczukin:
Thank you for the help @tianon. Looking forward to have this fix released into Docker Hub! :)

tianon:
Thank you! Feel free to follow along from here over at https://github.com/docker-library/official-images/pull/6403

norrs:
On behalf of the community and probably tons og gitlab dind users, thanks for the quick delivery and hard work @tmaczukin and @tianon !

PR #179, opened 2019-08-02
Restrict channel aliases to 19.03

### LINKED ISSUE #178: Use of the "stable" images tags
Hi everybody !

I'm opening this issue after suffering from two recent behaviour changes in the `dind` images (#175  and #170).

This happened despite using `stable` and `dind-stable` tags.

As far as I can tell, there is no difference between the following pairs of tags : 
* `stable` and `latest`
* `stable-dind` and `dind`
* `stable-dind-rootless` and `dind-rootless`
* `stable-git` and `git`

I'm a bit confused by this behaviour, because I would expect `stable` versions to lag behind cutting-edge versions, so that people specifying `stable` would have … well, more stable behaviour (some issues fixed before reaching stable, more time to migrate for important changes, etc…)

So, is there something I'm missing ? If not, can we think about setting up something like that ?

Of course, the exact workflow would need to be discussed, since this is a non-trivial question, but here are several ideas :
* Keeping `stable` one minor version older than the corresponding cutting-edge tag
* Delaying non-security issues by a week to catch most issues

tianon:
This will remove the "test" and "stable" aliases for whatever major release comes after 19.03 (so once the next release comes out, "stable" will become a stale alias).

I'm not 100% sure this is the right answer, but the "stable" alias has definitely been a huge source of confusion very recently (and it doesn't have much meaning anymore anyways given the death of the "edge" channel).

Closes #178

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -122,12 +122,12 @@ for version in "${versions[@]}"; do
 		${aliases[$version]:-}
 	)
 	if [ -z "${latestChannelRelease[$channel]:-}" ]; then
-		versionAliases+=( "$channel" )
+		if [ "$rcVersion" = '19.03' ]; then versionAliases+=( "$channel" ); fi # 19.03 is the last release to include any "channel" aliases
 		latestChannelRelease[$channel]="$version"
 	fi
 	# every release goes into the "test" channel, so the biggest numbered release wins (RC or not)
 	if [ -z "${latestChannelRelease['test']:-}" ]; then
-		versionAliases+=( 'test' )
+		if [ "$rcVersion" = '19.03' ]; then versionAliases+=( 'test' ); fi # 19.03 is the last release to include any "channel" aliases
 		latestChannelRelease['test']="$version"
 	fi
 	if [ "$version" = "$rcVersion" ] && [ -z "${latestChannelRelease['latest']:-}" ]; then


tianon:
@thaJeztah would you mind sharing any thoughts you might have on this? :pray: :heart:

thaJeztah:
hm, right, so isn't the confusion only there at moments when there's no beta/rc?

If there's a beta available, then `test` and `stable` are different.

I agree that "stable" has lost some of its meaning with "edge" no longer being a thing (just wondering if it'll break people); perhaps we should also look at stats, and see how popular the various tag permutations are

tianon:
The confusion I've seen recently has entirely been around the word `stable` -- believing it to somehow mean the tag will never change, or something?  Or that it will somehow be immune to backwards incompatibility?  I don't pretend to understand the full depth of the confusion, but I do understand the sentiment that `stable` isn't necessary a "stable" tag (and that it's really just the latest release of Docker, which is exactly what `latest` is/means).

I don't have any access to per-tag stats (and have been told that's not something we can realistically get), but if you've got them that would certainly be helpful. :sweat_smile:

tianon:
FWIW, the "breaking people" argument is why I think we should leave it around for 19.03, and once whatever's next comes out and 19.03 is no longer considered the latest "stable", the tag will simply drop away naturally (and become a stale "unsupported" tag).

luis12614:
Yes

tianon:
More cases of folks seeming to misunderstand what is meant by the "stable" tag: https://github.com/docker-library/docker/issues/240 :disappointed:

PR #191, opened 2019-09-12
introduce DOCKER_REMOUNT_SYS_RW (allow Docker-in-Kata without --privileged)

AkihiroSuda:
Docker-in-Kata can be launched with `--privileged` but it ruins the benefit of Kata because it mounts `/dev` from the host.

Now Docker-in-Kata can be launched without `--privileged`:

```console
$ docker run \
  --runtime kata \
  -e DOCKER_REMOUNT_SYS_RW=1 \
  --cap-add all \
  --security-opt seccomp=unconfined \
  --security-opt systempaths=unconfined \
  docker:dind
```

Tested with Kata Containers 1.8.0
(1.8.1 is broken: https://github.com/kata-containers/runtime/issues/2047)

Alternative to https://github.com/moby/moby/pull/39702

### DIFF
--- a/19.03-rc/dind/dockerd-entrypoint.sh
+++ b/19.03-rc/dind/dockerd-entrypoint.sh
@@ -88,10 +88,30 @@ _tls_generate_certs() {
 		openssl verify -CAfile "$dir/client/ca.pem" "$dir/client/cert.pem"
 	fi
 }
+_remount_sys_rw(){
+	mount -o remount,rw /sys
+	mount -o remount,rw /sys/fs/cgroup
+	grep "cgroup /sys/fs/cgroup" /proc/1/mounts | \
+		while read l; do
+			path=$(echo $l | cut -d " " -f 2)
+			opt=$(echo $l | cut -d " " -f 4)
+			umount $path
+			mount -t cgroup -o $(echo $opt | sed -e s/ro/rw/g) none $path
+		done
+}
 
 # no arguments passed
 # or first arg is `-f` or `--some-option`
 if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
+	if [ -n "${DOCKER_REMOUNT_SYS_RW:-}" ]; then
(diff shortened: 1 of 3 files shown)


thaJeztah:
Does this change also allow DIND on non-kata runtimes? 
Perhaps this would be a good example for inclusion in the docs for this image 🤔  (or would it be too specific if it's only for kata containers?)

> --cap-add all

Curious; do we know if a more limited set of caps works?

AkihiroSuda:
> Does this change also allow DIND on non-kata runtimes?

This doesn't seem to work on runc, however, even if it worked with runc, it doesn't make sense, because `systempaths=unconfined` allows container breakout (which is not problematic for Kata as long as the host devices are not mounted)

> Curious; do we know if a more limited set of caps works?

Probably we can drop some capabilities related to power management and Smack stuffs, but I don't see any benefit because anyway we need `systempaths=unconfined`

AkihiroSuda:
@tianon 

Do you think this should be moved to `hack/dind` in the moby repo?

tianon:
I do think this functionality makes more sense there, yes.

Do you think there's anything that can be done to auto-detect the "bad" state to some degree?

You can see the old "cgroups mounting" implementation in the commit which deleted it, which might be useful for reference: https://github.com/moby/moby/commit/81aa1b507f51901eafcfaad70a656da376cf937d#diff-fbcad9c68dbfcabbf5f6bf87019cc571

As a perhaps cleaner (in this case, meaning more generic) idea, is it potentially possible to detect that the `/sys/fs/cgroup` mounts are insufficient and simply remove them entirely, creating a whole new mount hierarchy to replace them with appropriate mounts? (which is roughly what that old code did)

https://github.com/tianon/cgroupfs-mount/blob/481f8b684fd67656f9d4c79828b48e24b6f9f321/cgroupfs-mount is probably another OK example, although it might have more edge cases (often used by folks/distributions without `systemd` automatically mounting a reasonable hierarchy).

This is also further complicated by cgroups v2, right?

AkihiroSuda:
Kata maintainers seems preferring https://github.com/moby/moby/pull/39702

PR #203, opened 2019-11-15
update RootlessKit to v0.7.0

AkihiroSuda:
Corresponds to Docker v19.03.5

https://github.com/docker/engine/blob/v19.03.5/hack/dockerfile/install/rootlesskit.installer


tianon:
Nice! Thanks!

PR #213, opened 2020-01-02
Change 4196-bit key size to 4096 bits

### LINKED ISSUE #212: 4196-bit key size instead of 4096
Is there a particular reason that a 4196-bit key is generated rather than a 4096-bit key? That is a rather unusual key size.
https://github.com/docker-library/docker/blob/9c726fa98bc90dec3d2b0dcd3fa00329b1967b98/dockerd-entrypoint.sh#L6

oakwhiz:
Closes #212


tianon:
Hah, great catch, thanks!

PR #220, opened 2020-03-09
Update rootlesskit to 0.9.1

tianon:
Also, update entrypoint for a few adjustments made upstream:

- https://github.com/moby/moby/pull/40635: launch rootlesskit with --propagation=rslave
- https://github.com/moby/moby/pull/39840: harden slirp4netns with mount namespace and seccomp

I was going to include https://github.com/moby/moby/pull/40406 as well, but it doesn't appear that https://github.com/moby/moby/commit/3518383ed990202d93e5458782d2c975c48ececd is in the released versions of Docker yet (and the additional argument should be reasonably harmless).

### DIFF
--- a/19.03-rc/dind-rootless/Dockerfile
+++ b/19.03-rc/dind-rootless/Dockerfile
@@ -47,7 +47,7 @@ RUN set -eux; \
 	vpnkit --version
 
 # https://github.com/rootless-containers/rootlesskit/releases
-ENV ROOTLESSKIT_VERSION 0.7.1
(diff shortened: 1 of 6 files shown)


tianon:
(cc @AkihiroSuda :heart:)

tianon:
So we should ditch that change here?

AkihiroSuda:
I think it is better. Also, `--slirp4netns-` flags can be removed unless we are going to add slirp4netns into the image.

tianon:
Ok, updated with just the version bump. :+1:

PR #235, opened 2020-05-18
Add gitpod config

victorvw:
this commit adds support for Gitpod.io, a free automated
dev environment that makes contributing and generally working on GitHub
projects much easier. It allows anyone to start a ready-to-code dev
environment for any branch, issue and pull request with a single click.


tianon:
While this looks neat, it's not something we use, nor something we want to maintain.  It also seems like something someone can/should maintain externally so they get their own preferred environment/settings?

PR #236, opened 2020-05-18
Support setting registry mirror via environment variables

Yannik:
This enables using registry mirrors in environments where the container command can not be modified.

### DIFF
--- a/19.03/dind/dockerd-entrypoint.sh
+++ b/19.03/dind/dockerd-entrypoint.sh
@@ -134,6 +134,14 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 	fi
 fi
 
+if [ "$DOCKER_REGISTRY_MIRROR" ]; then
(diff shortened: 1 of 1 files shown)


tianon:
Duplicate of #62.

What environment are you in where you cannot configure the command of the container to be run, but *can* modify/append environment variables?

Yannik:
@tianon gitlab with `disable_entrypoint_overwrite` set.

tianon:
Wouldn't that only block the ability to change `entrypoint:`, not `command:`?

tianon:
Confirmed via https://gitlab.hzdr.de/fwcc/gitlab-runner/commit/040a84d1050af1f5fd52b46339e2a0d591994a66#7a056729f6a9c4b323621c1ee3ee2894be489989 -- when `disable_entrypoint_overwrite` is set to `true`, it blocks the setting of `entrypoint:`, but not `command:` (which is IMO very weird, but means this isn't needed since `command:` is enough to add `--registry-mirror`). :+1:

PR #257, opened 2020-09-24
Add github container registry

fdns:


tianon:
I think you've probably filed this against the wrong repository?

yosifkit:
This git repository uses Docker Hub for the built images.

PR #258, opened 2020-10-05
Explicitly set --oom-score-adjust to -500

thaJeztah:
relates to https://github.com/moby/moby/pull/41528

Currently, dockerd sets a "-500" oom-score-adjust for itself by default.

We're removing this default, and instead:

- for dockerd running as systemd unit, set the oom-score through systemd
- when manually running dockerd, require users to explicitly set a score

This patch is in preparation of those changes.

### DIFF
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -118,6 +118,7 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 		set -- dockerd \
 			--host="$dockerSocket" \
 			--host=tcp://0.0.0.0:2376 \
+			--oom-score-adjust=-500 \
 			--tlsverify \
 			--tlscacert "$DOCKER_TLS_CERTDIR/server/ca.pem" \
 			--tlscert "$DOCKER_TLS_CERTDIR/server/cert.pem" \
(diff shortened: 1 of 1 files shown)


tianon:
Hmm, given there are a number of cases where neither of these lines will be used (such as `docker run ... docker:dind dockerd ...`), I think it makes sense to document this as a recommended parameter in https://github.com/docker-library/docs/blob/master/docker/content.md instead (with a link to https://github.com/moby/moby/pull/41528).  Thoughts?

thaJeztah:
Good one, yeah, I was wondering what would be best; keep a default, or don't set a default at all.

Given that it's running inside a container, users could even set the score of the container itself I guess 🤔

thaJeztah:
@cpuguy83 @AkihiroSuda wdyt?

cpuguy83:
dind is also not likely something that is super critical. The reason we set oomscore on dockerd is because it is typically used as a critical system process.

tianon:
Agreed -- I really like the idea of documenting that it should be set on the container instead, especially since that nicely mirrors setting it in the systemd service.  Maybe there are also other parameters in the systemd service we should document that users should set on their containers (`LimitNOFILE`, etc)?

thaJeztah:
closing in favour of https://github.com/docker-library/docs/pull/1821

PR #266, opened 2020-10-27
rootless: set ENV DOCKER_HOST=unix:///run/user/1000/docker.sock

AkihiroSuda:
This allows `docker exec $ID docker run ...` without setting `-H` manually.

### DIFF
--- a/19.03-rc/dind-rootless/Dockerfile
+++ b/19.03-rc/dind-rootless/Dockerfile
@@ -6,6 +6,8 @@
 
 FROM docker:19.03-rc-dind
 
+ARG ROOTLESS_UID=1000
+
 # busybox "ip" is insufficient:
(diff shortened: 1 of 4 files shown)


yosifkit:
Users are recommend to use `docker-entrypoint.sh` since it will automatically set this for more use cases (like running as a different user and TLS on the regular `dind`):

https://github.com/docker-library/docker/blob/d3041eb1bd203dcae86af9a121c7a8f6c8ef07a9/docker-entrypoint.sh#L26

See also https://github.com/docker-library/docker/pull/174#discussion_r309443928 (issue 174 is the only explanation in the ["Rootless" section](https://github.com/docker-library/docs/tree/1bf452e431173d225216a39e34f7b1d77e516e7c/docker#rootless) of the Docker Hub docs)

PR #269, opened 2020-11-14
Build rootlesskit in a companion container

vicamo:
Go build leaves some cache files in /root/.cache that takes up to 10MB space.

Signed-off-by: You-Sheng Yang <vicamo@gmail.com>

### DIFF
--- a/Dockerfile-dind-rootless.template
+++ b/Dockerfile-dind-rootless.template
@@ -1,3 +1,16 @@
+FROM golang:1.14-alpine{{ .alpine }} AS builder
+
+# https://github.com/rootless-containers/rootlesskit/releases
+ENV ROOTLESSKIT_VERSION 0.11.0
+
+RUN set -eux; \
+	wget -O rootlesskit.tgz "https://github.com/rootless-containers/rootlesskit/archive/v${ROOTLESSKIT_VERSION}.tar.gz"; \
+	mkdir -p "$GOPATH/src/github.com/rootless-containers/rootlesskit"; \
+	tar --extract --file rootlesskit.tgz --directory "$GOPATH/src/github.com/rootless-containers/rootlesskit" --strip-components 1; \
+	go build -o /usr/local/bin/rootlesskit github.com/rootless-containers/rootlesskit/cmd/rootlesskit; \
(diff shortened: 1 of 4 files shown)


vicamo:
```
bashbrew version v0.1.2
fatal: /usr/local/bin/rootlesskit*: '/usr/local/bin/rootlesskit*' is outside repository at '/home/runner/work/docker/docker'
+ strategy=
Error: Process completed with exit code 128.
```
Hmmm...

Dockerfile-dind-rootless.template
```diff
@@ -1,3 +1,16 @@
...
+# https://github.com/rootless-containers/rootlesskit/releases
+ENV ROOTLESSKIT_VERSION 0.11.0
+
+RUN set -eux; \
+	wget -O rootlesskit.tgz "https://github.com/rootless-containers/rootlesskit/archive/v${ROOTLESSKIT_VERSION}.tar.gz"; \
+	mkdir -p "$GOPATH/src/github.com/rootless-containers/rootlesskit"; \
+	tar --extract --file rootlesskit.tgz --directory "$GOPATH/src/github.com/rootless-containers/rootlesskit" --strip-components 1; \
+	go build -o /usr/local/bin/rootlesskit github.com/rootless-containers/rootlesskit/cmd/rootlesskit; \
+	go build -o /usr/local/bin/rootlesskit-docker-proxy github.com/rootless-containers/rootlesskit/cmd/rootlesskit-docker-proxy; \
+	rootlesskit --version
```
AkihiroSuda, line 12: Why don't we use the rootlesskit binary included in https://download.docker.com/linux/static/test/x86_64/docker-rootless-extras-20.10.0-beta1.tgz ? (I think this was intentional but I forgot the reason)

@tianon @yosifkit

tianon: I think it was probably just because rootless-extras for the released versions wasn't new enough?  If both 19.03.x (and 20.10.x) are new enough, I agree we could/should just pitch this entirely instead. :+1:

AkihiroSuda: Both should be new enough

vicamo:
Superseded by #273.

PR #273, opened 2020-11-23
Remove explicit rootlesskit version

tianon:
We were explicit about this previously because the version included with Docker wasn't new enough, and that's no longer the case.

Closes #269 (thanks for the prod, @vicamo!)


yosifkit:
An FYI, in `20.10.0-rc1`:
```console
$ rootlesskit --version
rootlesskit version 0.11.0
$ vpnkit --version
6bc1679a048d5d6837a37043e097f7b7cb710fdb
```
and in `19.03.13`:
```console
$ rootlesskit --version
rootlesskit version 0.10.0
$ vpnkit --version
67041ad2655038c5462b3466f89a1853f2b894e5
```

AkihiroSuda (approved):

PR #274, opened 2020-12-01
rootless: replace VPNKit with slirp4netns + support ARM

AkihiroSuda:
The rootless image was not built for ARM because VPNKit binary is not available for ARM.
This commit supports ARM by replacing VPNKit with slirp4netns.

slirp4netns also has several advantages over VPNkit (but not included in docker-rootless-extras archive because slirp4netns is GPL2):
- Faster than VPNKit, e.g. 0.60 Gbps vs 7.55 Gbps: https://github.com/rootless-containers/rootlesskit/tree/v0.11.1#network-drivers
- Can preserve src IP of incomming connections into `docker run -p` ports, by setting `$DOCKERD_ROOTLESS_ROOTLESSKIT_PORT_DRIVER` to "slirp4netns".
  (Otherwise incoming connections look like as if they were from localhost)
- Supports sandboxing slirp4netns itself using mount namespace and seccomp

### DIFF
--- a/19.03-rc/dind-rootless/Dockerfile
+++ b/19.03-rc/dind-rootless/Dockerfile
@@ -26,6 +26,15 @@ RUN set -eux; \
 		'x86_64') \
 			url='https://download.docker.com/linux/static/test/x86_64/docker-rootless-extras-19.03.13-beta2.tgz'; \
 			;; \
+		'armhf') \
+			url='https://download.docker.com/linux/static/test/armel/docker-rootless-extras-19.03.13-beta2.tgz'; \
+			;; \
+		'armv7') \
+			url='https://download.docker.com/linux/static/test/armhf/docker-rootless-extras-19.03.13-beta2.tgz'; \
+			;; \
+		'aarch64') \
+			url='https://download.docker.com/linux/static/test/aarch64/docker-rootless-extras-19.03.13-beta2.tgz'; \
+			;; \
 		*) echo >&2 "error: unsupported architecture ($apkArch)"; exit 1 ;; \
 	esac; \
 	\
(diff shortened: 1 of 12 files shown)


AkihiroSuda:
@yosifkit @tianon PTAL

AkihiroSuda:
rebased

AkihiroSuda:
@yosifkit @tianon @thaJeztah WDYT?

thaJeztah:
Was discussing this with @justincormack, and there were still some concerns; both w.r.t. the license, but also (from that conversation) because slirp4netns has had quite some CVE's. We could probably get VPNKit to be available for other architectures. That of course doesn't solve the performance issue, and we were wondering if we know what's causing the difference in performance; did you look into why that is, @AkihiroSuda ? (wondering if that can be fixed)

thaJeztah:
(is it only the jumbo frames, or is there another reason for the performance difference?)

AkihiroSuda:
> slirp4netns has had quite some CVE's

s/slirp4netns/libslirp/ 😛


Slirp4netns isolates itself from other processes using a mount namespace and seccomp, so basically unaffected by most of libslirp CVEs

AkihiroSuda:
> performance

Because of extra packet copy, VPNkit is slower even for small packets. I don't know about ocaml and how we can improve it.

https://github.com/rootless-containers/rootlesskit#network-drivers

thaJeztah:
> s/slirp4netns/libslirp/ 😛

Sorry, that's my mistake (mostly acting as "messenger" here; didn't want this PR to go silent 🤗)

> Because of extra packet copy, VPNkit is slower even for small packets. I don't know about ocaml and how we can improve it.

Thanks; @djs55 was also on that thread; he did want to have a look on the performance issue (if it was clear what was causing it); he has a lot on his plate currently though, so any help on that would of course be appreciated

AkihiroSuda:
As VPNkit is unaware of network namespaces, currently RootlessKit need to copy Ethernet packet data from VPNkit socket outside netNS to TAP device inside netNS, and vice versa. https://github.com/rootless-containers/rootlesskit/blob/7d4b61b7e0939e63d2d550139ee0ee0a96081b07/pkg/network/vpnkit/vpnkit.go#L190-L218

If VPNkit could this by itself (as in slirp4netns), we will be able to see huge throughput benefit.

AkihiroSuda:
@thaJeztah @justincormack 
Can we include slirp4netns at least for non-AMD64 archs? Currently, the dind image does not work on non-AMD64 hosts.

djs55:
In case it's useful I've built a amd64+arm64 image: https://hub.docker.com/layers/djs55/vpnkit/7f0eff0dd99b576c5474de53b4454a157c642834/images/sha256-81b2b42c9be1c14882e762a1ff2c55e210e7786b3a9d4844dc06d38436af79f0?context=explore . This doesn't address the performance problem though.

AkihiroSuda:
Thanks @djs55. Opened PR for adding VPNkit to arm64: https://github.com/moby/moby/pull/42067

PR #279, opened 2020-12-11
Update ENV syntax

J0WI:
https://github.com/docker/cli/pull/2743


tianon:
I really don't see a compelling reason for the churn (as I noted on that PR). :confused:

J0WI:
The only reason is to avoid deprecation warnings

tianon:
Closing since this isn't a change we intend to make (and as I commented on that upstream PR, I don't understand the deprecation to begin with :confused:).

PR #262, opened 2020-10-09
Add initial jq-based templating engine

yosifkit:
Based on https://github.com/docker-library/php/pull/1052 and related.  Also finally drops `windows` support from scripts.

```diff
$ diff -u ../official-images/library/docker <(./generate-stackbrew-library.sh)
--- ../official-images/library/docker	2020-09-17 15:37:41.373613739 -0700
+++ /dev/fd/63	2020-10-09 16:33:41.320816453 -0700
@@ -1,4 +1,4 @@
-# this file is generated via https://github.com/docker-library/docker/blob/606c63960a4845af7077721eb3900c706f5d0c5e/generate-stackbrew-library.sh
+# this file is generated via https://github.com/docker-library/docker/blob/dfced3292cb8cddaccac8365c4754a556f349809/generate-stackbrew-library.sh
 
 Maintainers: Tianon Gravi <tianon@dockerproject.org> (@tianon),
              Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
@@ -6,20 +6,20 @@
 
 Tags: 19.03.13, 19.03, 19, stable, test, latest
 Architectures: amd64, arm32v6, arm32v7, arm64v8
-GitCommit: 66c83a536259775694aa5d316426f3ff3ddcda6e
+GitCommit: dfced3292cb8cddaccac8365c4754a556f349809
 Directory: 19.03
 
 Tags: 19.03.13-dind, 19.03-dind, 19-dind, stable-dind, test-dind, dind
 Architectures: amd64, arm32v6, arm32v7, arm64v8
-GitCommit: 93c19dff54de9876f736b3aa8026be5ad5987388
+GitCommit: dfced3292cb8cddaccac8365c4754a556f349809
 Directory: 19.03/dind
 
 Tags: 19.03.13-dind-rootless, 19.03-dind-rootless, 19-dind-rootless, stable-dind-rootless, test-dind-rootless, dind-rootless
 Architectures: amd64
-GitCommit: 399f205733e0162fe6f6529fad8f2231b843606e
+GitCommit: dfced3292cb8cddaccac8365c4754a556f349809
 Directory: 19.03/dind-rootless
 
 Tags: 19.03.13-git, 19.03-git, 19-git, stable-git, test-git, git
 Architectures: amd64, arm32v6, arm32v7, arm64v8
-GitCommit: 7a67842e7ff12c1426ae6a67ac1b7a701b51f3df
+GitCommit: dfced3292cb8cddaccac8365c4754a556f349809
[...]

PR #311, opened 2021-06-01
Windows update

krystofwoldrich:
Hi guys,
I've needed to use docker on windows image and I've noticed the one in the repo is quite a bit old now. 

## Changes
- So I've updated the windows base image in the template
- and explicitly added windows version based on current ltsc2019.

Do you think it is welcomed change? 

I think the template would deserve the update.
And about the explicite file, I'm not sure if it won't collide with the ci here in the repo.

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -55,6 +55,15 @@ versionArches() {
 		selector='rootlessExtrasUrl'
 	fi
 
+	if [[ "$variant" = windows/* ]]; then
+		version="$version" jq -r '
+			.[env.version].arches
+			| keys[]
+			| select(startswith("windows-"))
+		' versions.json | sort
+		return
+	fi
+
 	local parent parentArches
 	parent="$(awk 'toupper($1) == "FROM" { print $2 }' "$version/Dockerfile")"
 	parentArches="${parentRepoToArches[$parent]:-}"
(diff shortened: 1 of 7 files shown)


tianon:
Whoa, these were originally removed because https://download.docker.com/win/static/stable/x86_64/ stopped having new builds, but it looks like that's changed??  @thaJeztah is that on purpose and going to stay true?? :pray:

thaJeztah:
> Whoa, these were originally removed because https://download.docker.com/win/static/stable/x86_64/ stopped having new builds, but it looks like that's changed?? @thaJeztah is that on purpose and going to stay true?? 🙏

Ah! You spotted them ☺️😁

Yes, Windows builds were disabled at some point because of `<reasons>`. Those reasons were no longer there, so when working on some changes in our release pipeline, we decided to "yolo! let's enable them again".

As to; will they stay? We have no intent to disable them again, but I must admit that I haven't tried the builds myself; working on a 20.10.7 test release right now; if those packages appear, give them a spin as well, and I'd say; go for it

tianon:
> Yes, Windows builds were disabled at some point because of `<reasons>`. Those reasons were no longer there, so when working on some changes in our release pipeline, we decided to "yolo! let's enable them again".
> 
> As to; will they stay? We have no intent to disable them again, but I must admit that I haven't tried the builds myself; working on a 20.10.7 test release right now; if those packages appear, give them a spin as well, and I'd say; go for it

Ahh, makes sense -- we'll give it a go!

> I think the template would deserve the update.
> And about the explicite file, I'm not sure if it won't collide with the ci here in the repo.

Yeah, this is a good start, although we also need to resurrect some of the bits that were removed in https://github.com/docker-library/docker/pull/262 (honestly, the template was an oversight there :joy:).

I'll take over from here to get this updated with the missing bits. :sweat_smile:

krystofwoldrich:
> > Yes, Windows builds were disabled at some point because of `<reasons>`. Those reasons were no longer there, so when working on some changes in our release pipeline, we decided to "yolo! let's enable them again".
> > As to; will they stay? We have no intent to disable them again, but I must admit that I haven't tried the builds myself; working on a 20.10.7 test release right now; if those packages appear, give them a spin as well, and I'd say; go for it
> 
> Ahh, makes sense -- we'll give it a go!
> 
> > I think the template would deserve the update.
> > And about the explicite file, I'm not sure if it won't collide with the ci here in the repo.
> 
> Yeah, this is a good start, although we also need to resurrect some of the bits that were removed in #262 (honestly, the template was an oversight there 😂).
> 
> I'll take over from here to get this updated with the missing bits. 😅

Haha, what a coincidence. I’m happy that now it will be back.

generate-stackbrew-library.sh
```diff
@@ -154,21 +163,32 @@ for version; do
...
+		dir="$version${v:+/$v}"
 		[ -f "$dir/Dockerfile" ] || continue
 
 		commit="$(dirCommit "$dir")"
 
+		variant="$(basename "$v")"
 		variantAliases=( "${versionAliases[@]/%/${variant:+-$variant}}" )
 		variantAliases=( "${variantAliases[@]//latest-/}" )
 
+		sharedTags=()
```
yosifkit, line 176: Do we want shared tags between Windows & Linux too? (like python)

tianon: We didn't previously, which is why I didn't here -- I think that was a conscious choice due to the massive differences in their ability to be functional/useful on the different platforms.

yosifkit (approved):

PR #317, opened 2021-07-14
update alpine to 3.14

AkihiroSuda:
Required for https://github.com/docker-library/docker/pull/265

### DIFF
--- a/20.10-rc/Dockerfile
+++ b/20.10-rc/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM alpine:3.13
+FROM alpine:3.14
 
 RUN apk add --no-cache \
 		ca-certificates \
--- a/20.10/Dockerfile
+++ b/20.10/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM alpine:3.13
+FROM alpine:3.14
 
 RUN apk add --no-cache \
 		ca-certificates \
--- a/versions.json
+++ b/versions.json
@@ -1,6 +1,6 @@
 {
   "20.10": {
-    "alpine": "3.13",
+    "alpine": "3.14",
     "arches": {
       "amd64": {
         "dockerUrl": "https://download.docker.com/linux/static/stable/x86_64/docker-20.10.8.tgz",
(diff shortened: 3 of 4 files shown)


tianon:
This is definitely going to cause issues/headaches for users :disappointed:

(https://github.com/docker-library/postgres/issues/864, https://github.com/docker-library/memcached/issues/68, https://github.com/docker-library/httpd/issues/185 for a few examples)

tianon:
@tmaczukin are you still a good person to ping on the GitLab side?  I'm wanting to give you a heads up that this PR is coming and has a high chance of breaking folks if the host's Docker is not new enough (and that #327 is going to provide a minimal escape hatch for users who need to pin to Alpine 3.13 because they can't upgrade yet).

tmaczukin:
> @tmaczukin are you still a good person to ping on the GitLab side? I'm wanting to give you a heads up that this PR is coming and has a high chance of breaking folks if the host's Docker is not new enough (and that #327 is going to provide a minimal escape hatch for users who need to pin to Alpine 3.13 because they can't upgrade yet).

Hey @tianon, thanks for the notification!

We've already found Alpine 3.14 to be problematic with older Docker Engine versions. For GitLab.com instance runners that we provide it's no more a problem (so also upgrade of `docker` and `docker:dind` images shouldn't be) as we've upgraded the underlying OS along with Docker. But self-hosted GitLab Runner users will need to handle this on their own.

I'm not sure we can do anything to help this apart of informing about the upcoming upgrade. After the last problem in 2019 we've wrote a blog post and updated our documentation to make it clear why depending on floating versions like `:latest` or `:dind` is a problem and that it's highly suggested to use pined versions. But there are probably many users who still don't do that.

And after a quick look at the changes here, it seems that even the pined versions like `docker:19.03.9-dind` **will get updated** to a newer Alpine version. Which means that even the more careful/aware users may got affected, and this makes #327 really important to have as a way for a quick fix (updating OS or Docker Engine on user's machines may be not as easy task).
 
Is there any plan when #327 and #317 (this one) would be merged? Is it possible to make the Alpine 3.13 based versions the default ones and add Alpine 3.14 version as an option for now and have some grace period before it will become the default?

To get ahead of potential user issues we will probably publish a blog post about the upcoming change. Knowing when it will be merged and having a simple workaround solution (like #327) would be highly appreciated :)

tianon:
We appreciate the reply!! :heart:

You helped prompt #329 (which should merge soon, and thus make at least all 19.x tags static/safe from here forward), and we've already merged #327 all the way through, so those `alpine3.13` aliases should be available now (or shortly).

Ideally we'd like to merge this ASAP, but I don't have a problem giving it a little more time to bake with #327 out for folks to preemptively switch over if you think that doing so would actually be effective.

AkihiroSuda:
rebased

tianon:
:eyes:

PR #324, opened 2021-08-03
Use full version in FROM lines

thaJeztah:
Perhaps there is a specific reason for not doing this, but when trying to
build the image locally, I made the mistake of not tagging a version without
the patch release, and ended up with a `-dind` image that was based on the
previous version;

    docker build -t docker:20.10.8 ./20.10/
    docker build -t docker:20.10.8-dind ./20.20/dind/

Using the full version makes sure that the base image is the expected (current)
version.

### DIFF
--- a/19.03-rc/dind-rootless/Dockerfile
+++ b/19.03-rc/dind-rootless/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM docker:19.03-rc-dind
+FROM docker:19.03.13-beta2-dind
 
 # busybox "ip" is insufficient:
 #   [rootlesskit:child ] error: executing [[ip tuntap add name tap0 mode tap] [ip link set tap0 address 02:50:00:00:00:01]]: exit status 1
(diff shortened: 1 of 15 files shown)


thaJeztah:
@tianon @yosifkit ptal; feel free to close if this makes no sense 🤗 😂 . I made this mistake when building, and was wondering if there was a reason for not using the full version; if there is; go ahead and close it 💥

tianon:
Mostly we avoid this because it creates some amount of conflation between the tags-generating code in `generate-stackbrew-library.sh` and the `Dockerfile` templates (whereas we know that `env.version` will always be one of the tag aliases for any given generated `Dockerfile`).

This also keeps the number of "changed lines" lower for each version bump (one total instead of one-per-`Dockerfile`).

think101 (approved):

tao12345666333:
> This also keeps the number of "changed lines" lower for each version bump (one total instead of one-per-`Dockerfile`).

I agree with @tianon 

Although "best practice" always tells us to use the full version number, I think there is an exception here. :cat:

tianon:
:heart:

Issue #338, opened 2021-12-15
Upgrade alpine version


Butzlabben:
Hi,

it would be really nice, if you could update the alpine version from 3.14 to 3.15

ghevge:
+1 Any ETA on this ? I am seeing an   make: /bin/sh: Operation not permitted error on one of my dind containers which seem to have been fixed  3.15

PR #339, opened 2021-12-15
update alpine to 3.15

tao12345666333:
resolve https://github.com/docker-library/docker/issues/338


tianon:
Nice, thanks @tao12345666333!

Issue #156, opened 2019-05-03
Include docker-app plugin with docker image


kinghuang:
*Copied from docker/app#532*

**Description**

Can the docker-app 0.8.0 CLI plugin be included with the [docker](https://hub.docker.com/_/docker/) 19.03 image (currently at docker:19.03.0-beta3)?

I previously asked for a docker-app image in docker/app#367, similar to the [docker/compose](https://hub.docker.com/r/docker/compose) image. But, with docker-app now functioning as a CLI plugin, it would be preferable to have it as part of the docker image.

**Steps to reproduce the issue:**

1. Run a container with the docker:19.03.0-beta3 image.
2. Attempt to use `docker app` commands.

**Describe the results you received:**

`docker app` commands aren't available. The docker-app CLI plugin is not included with the docker image.

**Describe the results you expected:**

Be able to use `docker app` commands.

**Additional information you deem important (e.g. issue happens only occasionally):**

There is currently no official image with Docker App, either as a plugin or as a standalone command. I need an image for CI jobs that use Docker App.

**Output of `docker version`:**

```
Client: Docker Engine - Community
 Version:           19.03.0-beta3
 API version:       1.40
 Go version:        go1.12.4
 Git commit:        c55e026
 Built:             Thu Apr 25 02:55:45 2019
 OS/Arch:           linux/amd64
 Experimental:      false

```

**Output of `docker-app version`:**

Docker App is not available in the docker image.

**Output of `docker info`:**

N/A

**Additional environment details (AWS, VirtualBox, physical, etc.):**

tianon:
This image faithfully publishes whatever bits Docker officially provides via https://download.docker.com/linux/static/ (installing necessary dependencies and some very short scripts to help with usage in minor ways).

tianon:
To put that another way, I don't think it makes sense for us to officially bundle any additional CLI (or otherwise) plugins that Docker's standard distribution does not include (since our main goal is to be a faithful representation of Docker's upstream bits).

kinghuang:
Ah, ok. I'll close this request and focus on asking for a docker-app image. Thanks!

ijc:
FWIW my reading of https://github.com/docker-library/docker/issues/156#issuecomment-489245797 is that we should (try and) arrange for plugins to be part of https://download.docker.com/linux/static/ and then it would be possible to include them in this image. @tianon did I understand you correctly?

tianon:
[...]

Issue #313, opened 2021-06-14
buildx is missing from docker:20-dind images


anapsix:
**Description**

Docker 20-dind images are shipped with `docker` cli which seems to be missing `buildx` entirely.. which means support for `buildx build --cache-from` is not available.
Launching `docker buildx build --help` on my Macbook shows `--cache-from` available. However launching the same in container instance of `docker:20-dind` does not show `--cache-from` as valid option to `buildx build`.

**Steps to reproduce the issue:**
```sh
$ docker pull docker:20.10.7-dind
$ docker run -it --rm -e DOCKER_CLI_EXPERIMENTAL=enabled docker:20.10.7-dind sh -c 'docker --help | grep buildx && echo SUCCESS: buildx is available || echo ERROR: buildx is not available'
$ docker run -it --rm docker:20.10.7-dind sh -c 'docker --version; docker buildx build --help | grep cache-from || echo "ERROR: buildx build --cache-from is not supported"'
```

**Describe the results you received:**
```
ERROR: buildx is not available
Docker version 20.10.7, build f0df350
ERROR: buildx build --cache-from is not supported
```

**Describe the results you expected:**
```
SUCCESS: buildx is available
Docker version 20.10.7, build f0df350
      --cache-from stringArray   External cache sources (eg. user/app:cache, type=local,src=path/to/dir)
```

**Additional information you deem important (e.g. issue happens only occasionally):**

**Output of `docker version`:**
from `docker:20.10.7-dind` docker image
```
Client:
 Version:           20.10.7
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        f0df350
 Built:             Wed Jun  2 11:51:04 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true
```

**Output of `docker info`:**
from `docker:20.10.7-dind` docker image
```
Client:
 Context:    default
 Debug Mode: false

Server:
ERROR: error during connect: Get http://docker:2375/v1.24/info: dial tcp: lookup docker on 192.168.65.5:53: no such host
errors pretty printing info
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**
Launching `docker run -it --rm docker:20.10.7-dind sh -c ...` on macOS 10.14.6 running `Docker Desktop 3.4.0 (3.4.0.5223)`

tianon:
Duplicate of #243, #156

anapsix:
[...]

Issue #243, opened 2020-07-13
Include buildx cli-plugins


BarnabyShearer:
Docker 19.03 includes the `buildx` CLI plugin by default. It would be very helpful if this image could match that.

tianon:
Unfortunately, this is a duplicate of #156

velcrine:
Why the headache to create and maintain a new dockerfile just for one copy command. Can't  docker host it?

Issue #42, opened 2017-02-21
docker-compose in docker


tinytelly:
I have docker-compose installed on the server like this:
```
$ docker-compose --version
docker-compose version 1.8.1, build 878cff1
$ which docker-compose
/usr/local/bin/docker-compose
```

I call my docker processes like this:
```
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock myservice:latest
```
inside the container I call a bash script that does docker things..and all works fine.
I also want to call docker-compose and have it use the docker-compose on the host in the same as docker does.

Not sure how to do this.....any ideas?

Thanks
Matt

tinytelly:
figured it out....here's the answer to help anyone that is trying to do the same thing.

Inside the bash script running in the container you run docker-compose as a container rather than try to use the docker-compose on the host.

Like this:
`docker run --rm  -v /var/run/docker.sock:/var/run/docker.sock -v /home:/home -w /home docker/compose:1.11.2 -f docker-compose-you-want-to-run.yml  up -d --no-recreate`

make sure you copy docker-compose-you-want-to-run.yml into /home first.

PR #361, opened 2022-05-06
Add "buildx" and "compose" CLI plugins (for supported architectures of each)

### LINKED ISSUE #303: DinD with docker-compose
After following the guide in https://hub.docker.com/_/docker/ I wrote this docker-compose.yaml that should setup the same as the shell commands (if I understood it correctly):

``` yaml
version: "3.8"

services:
  dind:
    image: docker:dind
    privileged: true
    environment:
      DOCKER_TLS_CERTDIR: /certs
    networks:
      - docker
    volumes:
      - docker-certs-ca:/certs
      - docker-certs-client:/certs/client

  docker-client:
    image: docker
    environment:
      DOCKER_TLS_CERTDIR: /certs
    # Just loop at infinitum, we will get to this container by `docker exec`ing
    command: [sh, -c, "while sleep 1; do :; done"]
    networks:
      - docker
    depends_on:
      - dind
    volumes:
      - docker-certs-client:/certs/client:ro

volumes:
  docker-certs-ca:
  docker-certs-client:

networks:
  docker:
```

After running `docker-compose exec docker-client docker version` I got the following error, and I cannot connect to the dind docker from the docker-client:

```
Client: Docker Engine - Community
 Version:           20.10.5
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        55c4c88
 Built:             Tue Mar  2 20:14:11 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

Is there anything I'm missing in my docker-compose.yml?

Running the commands from the README manually I get the correct result

### LINKED ISSUE #355: Why doesn't `docker:latest` image contains the compose?
When I tried to run `docker compose up` I got `docker: 'compose' is not a docker command.` but in `docker:dind` always good. Maybe `compose` missed from `docker:latest` image?

tianon:
- Closes #303
- Closes #355
- Closes https://github.com/docker/roadmap/issues/322

(I don't actually have permission for this to auto-close https://github.com/docker/roadmap/issues/322, so maybe someone like @thaJeztah or @chris-crone can do the honors once we get this in.)

Refs:
- #18
- #30
- #42
- #156
- #243
- #313
- https://github.com/docker/compose/issues/8871
- https://github.com/docker/roadmap/issues/256
- https://github.com/docker/docker-ce-packaging/pull/309

As a warning, this does very severely bloat the size of the Docker images (something like 100MiB more), but given these are part of the "Docker Experience" I think it makes sense. :see_no_evil:

### DIFF
--- a/20.10/Dockerfile
+++ b/20.10/Dockerfile
@@ -55,6 +55,86 @@ RUN set -eux; \
 	dockerd --version; \
 	docker --version
 
+ENV DOCKER_BUILDX_VERSION 0.8.2
+RUN set -eux; \
+	apkArch="$(apk --print-arch)"; \
+	case "$apkArch" in \
+		'x86_64') \
+			url='https://github.com/docker/buildx/releases/download/v0.8.2/buildx-v0.8.2.linux-amd64'; \
+			sha256='c64de4f3c30f7a73ff9db637660c7aa0f00234368105b0a09fa8e24eebe910c3'; \
+			;; \
+		'armhf') \
+			url='https://github.com/docker/buildx/releases/download/v0.8.2/buildx-v0.8.2.linux-arm-v6'; \
+			sha256='d0e5d19cd67ea7a351e3bfe1de96f3d583a5b80f1bbadd61f7adcd61b147e5f5'; \
+			;; \
+		'armv7') \
+			url='https://github.com/docker/buildx/releases/download/v0.8.2/buildx-v0.8.2.linux-arm-v7'; \
+			sha256='b5bb1e28e9413a75b2600955c486870aafd234f69953601eecc3664bd3af7463'; \
+			;; \
+		'aarch64') \
+			url='https://github.com/docker/buildx/releases/download/v0.8.2/buildx-v0.8.2.linux-arm64'; \
+			sha256='304d3d9822c75f98ad9cf57f0c234bcf326bbb96d791d551728cadd72a7a377f'; \
+			;; \
+		'ppc64le') \
+			url='https://github.com/docker/buildx/releases/download/v0.8.2/buildx-v0.8.2.linux-ppc64le'; \
+			sha256='32b317d86c700d920468f162f93ae2282777da556ee49b4329f6c72ee2b11b85'; \
+			;; \
+		'riscv64') \
+			url='https://github.com/docker/buildx/releases/download/v0.8.2/buildx-v0.8.2.linux-riscv64'; \
+			sha256='76d5fcf92ffa31b3e470d8ec1ab11f7b6997729e5c94d543fec765ad79ad0630'; \
+			;; \
+		's390x') \
+			url='https://github.com/docker/buildx/releases/download/v0.8.2/buildx-v0.8.2.linux-s390x'; \
+			sha256='ec4bb6f271f38dca5a377a70be24ee2108a85f6e6ba511ad3b805c4f1602a0d2'; \
+			;; \
+		*) echo >&2 "warning: unsupported buildx architecture ($apkArch); skipping"; exit 0 ;; \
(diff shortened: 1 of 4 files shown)


tao12345666333:
👍 great!

By the way, can we provide a tag like non-plugin, which means that plugin is not included.

Because in our actual scenario, DinD is heavily used in CI without the need to use plugins.  It would be nice to not increase the size of the image. Thanks

tianon:
:see_no_evil: that's exactly the type of request I've been hoping to avoid all this time :sweat_smile:

If anything, I'd love to maintain _less_ variations of this image than more. :innocent:

I've also been _very_ strongly considering moving from Alpine to Debian or Ubuntu for the next (22.xx) release, since the experience of using Docker-Inc-provided Docker builds on both of those platforms is much more well-supported than Alpine is. :disappointed:

johnthagen:
> I've also been very strongly considering moving from Alpine to Debian or Ubuntu for the next (22.xx) release

As a drive by comment, I too will say that I've hit rough edged working with the Alpine-derived Docker images. If you don't need to do any changes they work well, but the moment you want to add some scripts or anything else to image Alpine is harder to use.

Also big ❤️ for getting Compose v2 into an official image.

tianon:
With the implementation currently in this PR, new versions of Compose will be integrated as they're released (so you'd have to pin to an image digest if you want something that doesn't change).

wedi:
I am used to Alpine from the times when there was no great slim alternative and I have been sticking to it to copy and paste proven recipes. Now that we have, for example, an Ubuntu image with a footprint below 30 MB, I think that in principle it would be fine to switch. However, please keep in mind how many people use these images. Making the switch will easily require thousands of developer hours in downstream projects and Pipelines. So it would be nice if you kept it as long as it doesn't become a major drain on your time.

wedi:
Once merged this would allow to finally resolve https://github.com/docker-library/docker/issues/156.

tianon:
Just to record this somewhere, I did some testing on combining layers here to see if we get some compression efficiencies on transport (h/t @chris-crone), and the answer is a big resounding "no" -- building this as-is, the total "compressed size" of all layers is 89.47MB.  If I combine both plugins layers into a single one, it's the same.  Even if I combine the full Docker/containerd binaries layer as well, it's still exactly the same. :grimacing:

(So my preference is to continue to leave them separate.)

tianon:
I guess this is also a good place to note that the comparison to "current" from that is ~66.99MB, so for transport, this is not a _huge_ increase.

chris-crone:
> Just to record this somewhere, I did some testing on combining layers here to see if we get some compression efficiencies on transport (h/t @chris-crone), and the answer is a big resounding "no" -- building this as-is, the total "compressed size" of all layers is 89.47MB. If I combine both plugins layers into a single one, it's the same. Even if I combine the full Docker/containerd binaries layer as well, it's still exactly the same. 😬
> 
> (So my preference is to continue to leave them separate.)

I was so sure that gzip would make the duplicated Go runtime go away but alas 😢

johnthagen:
Now that Compose v2 is included in the [`docker`](https://hub.docker.com/_/docker) image, should [`docker/compose` (v1)](https://hub.docker.com/r/docker/compose) be updated with a notice that it deprecated?

johnthagen:
@tianon I don't know if it's moving up to Compose v2 or buildx or some combination, but when we moved our CI that runs `docker/compose:1.29.2`/`docker-compose build --parallel` -> `docker`/`docker compose build` our compose project (multiple multi-stage Dockerfiles) built **twice** as fast. ❤️ 

Thanks!

PR #369, opened 2022-06-16
Add explicit "cli" aliases and move "latest" to "dind" in 22.06+

tianon:
cc @thaJeztah


tianon:
<details>
<summary>Diff:</summary>

```diff
--- /dev/fd/63	2022-06-16 14:12:16.722850613 -0700
+++ /dev/fd/62	2022-06-16 14:12:16.722850613 -0700
@@ -1,24 +1,24 @@
 Maintainers: Tianon Gravi <tianon@dockerproject.org> (@tianon), Joseph Ferguson <yosifkit@gmail.com> (@yosifkit)
 GitRepo: https://github.com/docker-library/docker.git
 
-Tags: 22.06.0-beta.0, 22.06-rc, rc, 22.06.0-beta.0-alpine3.16
+Tags: 22.06.0-beta.0-cli, 22.06-rc-cli, rc-cli, 22.06.0-beta.0-cli-alpine3.16
 Architectures: amd64, arm64v8
-GitCommit: 4c1b73e7046422ca1339b76935c2e2f4251b9c20
-Directory: 22.06-rc
+GitCommit: b6963b10e947cc8f756d8910833dca36e8ed5804
+Directory: 22.06-rc/cli
 
-Tags: 22.06.0-beta.0-dind, 22.06-rc-dind, rc-dind, 22.06.0-beta.0-dind-alpine3.16
+Tags: 22.06.0-beta.0-dind, 22.06-rc-dind, rc-dind, 22.06.0-beta.0-dind-alpine3.16, 22.06.0-beta.0, 22.06-rc, rc, 22.06.0-beta.0-alpine3.16
 Architectures: amd64, arm64v8
-GitCommit: 4c1b73e7046422ca1339b76935c2e2f4251b9c20
+GitCommit: b6963b10e947cc8f756d8910833dca36e8ed5804
 Directory: 22.06-rc/dind
 
 Tags: 22.06.0-beta.0-dind-rootless, 22.06-rc-dind-rootless, rc-dind-rootless
 Architectures: amd64, arm64v8
-GitCommit: 4c1b73e7046422ca1339b76935c2e2f4251b9c20
+GitCommit: b4ebf24fc840e7f223c9e7ab361a02fa2ffd2e3e
 Directory: 22.06-rc/dind-rootless
 
 Tags: 22.06.0-beta.0-git, 22.06-rc-git, rc-git
 Architectures: amd64, arm64v8
-GitCommit: 4c1b73e7046422ca1339b76935c2e2f4251b9c20
+GitCommit: b6963b10e947cc8f756d8910833dca36e8ed5804
 Directory: 22.06-rc/git
 
 Tags: 22.06.0-beta.0-windowsservercore-ltsc2022, 22.06-rc-windowsservercore-ltsc2022, rc-windowsservercore-ltsc2022
@@ -35,24 +35,24 @@
 Directory: 22.06-rc/windows/windowsservercore-1809
 Constraints: windowsservercore-1809
 
-Tags: 20.10.17, 20.10, 20, latest, 20.10.17-alpine3.16
+Tags: 20.10.17-cli, 20.10-cli, 20-cli, cli, 20.10.17-cli-alpine3.16, 20.10.17, 20.10, 20, latest, 20.10.17-alpine3.16
 Architectures: amd64, arm64v8
-GitCommit: 034e1d03eaf2c35ce4a21c4a4b621e2d2174b62e
-Directory: 20.10
+GitCommit: b6963b10e947cc8f756d8910833dca36e8ed5804
+Directory: 20.10/cli
 
 Tags: 20.10.17-dind, 20.10-dind, 20-dind, dind, 20.10.17-dind-alpine3.16
 Architectures: amd64, arm64v8
-GitCommit: 0efba9e3cd4537de89ba54de2ad8acc5e3b1759f
+GitCommit: b6963b10e947cc8f756d8910833dca36e8ed5804
 Directory: 20.10/dind
 
 Tags: 20.10.17-dind-rootless, 20.10-dind-rootless, 20-dind-rootless, dind-rootless
 Architectures: amd64, arm64v8
-GitCommit: 034e1d03eaf2c35ce4a21c4a4b621e2d2174b62e
+GitCommit: b4ebf24fc840e7f223c9e7ab361a02fa2ffd2e3e
 Directory: 20.10/dind-rootless
 
 Tags: 20.10.17-git, 20.10-git, 20-git, git
 Architectures: amd64, arm64v8
-GitCommit: 387e351394bfad74bceebf8303c6c8e39c3d4ed4
+GitCommit: b6963b10e947cc8f756d8910833dca36e8ed5804
 Directory: 20.10/git
 
 Tags: 20.10.17-windowsservercore-ltsc2022, 20.10-windowsservercore-ltsc2022, 20-windowsservercore-ltsc2022, windowsservercore-ltsc2022
```

</details>

tianon:
This is probably more interesting if/when we get that cli/engine tarball split so that we can actually have a meaningfully smaller cli variant :thinking:

thaJeztah:
> This is probably more interesting if/when we get that cli/engine tarball split so that we can actually have a meaningfully smaller cli variant 

Would it be an option to already remove the non-cli binaries (containerd, dockerd, runs) as part of the Dockerfile?

tao12345666333:
I think if we call it as "cli", maybe we should only include cli binary, right?

tianon:
@tao12345666333 indeed! :sweat_smile: :heart:

That was always the _intent_ of this variant, as hinted by this (added back in https://github.com/docker-library/docker/pull/64; five years ago :weary: :older_adult:):

https://github.com/docker-library/docker/blob/038fb86c696159554fb1e7f9f43a7d72de872748/docker-entrypoint.sh#L46-L59

However, Docker only publishing the single `docker.tgz` that includes everything has always made that a little bit annoying to accomplish cleanly.  However, I've got a new change I'll push to this branch shortly that finally does the redownload necessary to accomplish that (and hopefully it'll be split properly in some future version -- work is currently being done around that in https://github.com/docker/docker-ce-packaging/pull/665). :eyes:

tianon:
The justification for switching the default tag to `dind` is that it's both what most users actually want/use _and_ it still contains the CLI binary so shouldn't be too bad of a breaking change for anyone (and for anyone that it is, it's only in 22.06+ _and_ there's a new explicit `-cli` tag that more clearly communicates intent).

The justification for removing `dockerd` in the CLI tags is a little less strong, but is definitely in line with the intention of them all along (as evidenced by the warning they produce if you try to run `dockerd` in them).  If folks are running the daemon in the CLI tags today, they're likely doing some backflips to accomplish that which they could (and should) stop doing if they switch to the explicit `dind` tags instead. :+1:

yosifkit (approved):

Issue #397, opened 2023-02-03
dind rootless fails and fallback to unsecure non-tls when /certs/ca is mounted as volume


PierreBeucher:
When `docker:dind-rootless` starts and a volume is configured on `/certs/ca`, the daemon cannot generate TLS certificates and falls back to non-TLS with errors like:

```
dockerd  | genrsa: Can't open "/certs/ca/key.pem" for writing, Permission denied
dockerd  | Could not open file or uri for loading private key from /certs/ca/key.pem
dockerd  | 481B2FF9CC7F0000:error:16000069:STORE routines:ossl_store_get0_loader_int:unregistered scheme:crypto/store/store_register.c:237:scheme=file
dockerd  | 481B2FF9CC7F0000:error:80000002:system library:file_open:No such file or directory:providers/implementations/storemgmt/file_store.c:267:calling stat(/certs/ca/key.pem)
```

The daemon still starts but without TLS.

Minimal reproducible example with Compose:

```yaml
version: "3.9"

services:
  dockerd:
    image: docker:23.0.0-dind-rootless
    container_name: dockerd
    privileged: true
    volumes:
      - docker_certs_client:/certs/client
      - docker_certs_ca:/certs/ca

volumes:
  docker_certs_client:
  docker_certs_ca: 
```

A workaround is to mount `/certs` instead, but it's less secure  when sharing volume with other containers are CA data is also available to "client" containers.

yosifkit:
:thinking: the `openssl` command not causing it to stop earlier and just automatically disabling TLS is definitely a bug (https://github.com/docker-library/docker/pull/398).

As for using `-rootless` with multiple directories, I think they need to somehow be owned my the right user, but docker doesn't give any way to control that.

Have you tried this? I think the permissions we create on the directories to work with rootless only apply to `/certs` and `/certs/client`.
```console
    volumes:
      - docker_certs_client:/certs/client
      - docker_certs:/certs

volumes:
  docker_certs_client:
  docker_certs: 
```

PierreBeucher:
Thanks for the prompt answer! Indeed this does the trick, thanks. May it be worth mentioning on doc to avoid surpises?

PR #398, opened 2023-02-03
prevent tls being disabled if _tls_generate_certs fails

yosifkit:
If TLS is requested (via `DOCKER_TLS_CERTDIR`) and the directory is not writable, then it should not silently disable TLS.

Fixes https://github.com/docker-library/docker/issues/397


tianon (approved):
:facepalm: this is my bad - trying to be too clever :bow:

PR #417, opened 2023-03-31
Add 24.0.0-beta.1

tianon:
FYI @thaJeztah

### DIFF
--- a/24-rc/dind/Dockerfile
+++ b/24-rc/dind/Dockerfile
@@ -1,0 +1,89 @@
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "apply-templates.sh"
+#
+# PLEASE DO NOT EDIT IT DIRECTLY.
+#
+
+FROM docker:24-rc-cli
+
+# https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
+RUN set -eux; \
+	apk add --no-cache \
+		btrfs-progs \
+		e2fsprogs \
+		e2fsprogs-extra \
+		ip6tables \
+		iptables \
+		openssl \
+		shadow-uidmap \
+		xfsprogs \
+		xz \
+# pigz: https://github.com/moby/moby/pull/35697 (faster gzip implementation)
+		pigz \
+	; \
+# only install zfs if it's available for the current architecture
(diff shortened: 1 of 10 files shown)


24-rc/dind/Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+# https://github.com/docker/docker/tree/master/hack/dind
+ENV DIND_COMMIT 1f32e3c95d72a29b3eaacba156ed675dba976cb5
+
+RUN set -eux; \
+	wget -O /usr/local/bin/dind "https://raw.githubusercontent.com/docker/docker/${DIND_COMMIT}/hack/dind"; \
+	chmod +x /usr/local/bin/dind
+
+COPY dockerd-entrypoint.sh /usr/local/bin/
+
+VOLUME /var/lib/docker
```
thaJeztah, line 85: We should probably start thinking about containerd storage paths as well (where do we store images for containerd when running as child process?)

thaJeztah (commented):
LGTM! Thank you ❤️

thaJeztah:
And... it's ✅ green 🎉 /cc @yosifkit

24-rc/dind/Dockerfile
tianon: I agree, but I just tested and it's still under `/var/lib/docker` in this configuration (because of Docker's containerd configuration), so I think we're good.

tianon:
```console
/ # ls -l /var/lib/docker/containerd/daemon/
total 36
drwxr-xr-x    4 root     root          4096 Apr  1 00:20 io.containerd.content.v1.content
drwx------    2 root     root          4096 Apr  1 00:20 io.containerd.grpc.v1.introspection
drwx--x--x    2 root     root          4096 Apr  1 00:20 io.containerd.metadata.v1.bolt
drwx--x--x    2 root     root          4096 Apr  1 00:20 io.containerd.runtime.v1.linux
drwx--x--x    2 root     root          4096 Apr  1 00:20 io.containerd.runtime.v2.task
drwx------    2 root     root          4096 Apr  1 00:20 io.containerd.snapshotter.v1.btrfs
drwx------    3 root     root          4096 Apr  1 00:20 io.containerd.snapshotter.v1.native
drwx------    3 root     root          4096 Apr  1 00:20 io.containerd.snapshotter.v1.overlayfs
drwx------    2 root     root          4096 Apr  1 00:20 tmpmounts
```

PR #419, opened 2023-04-03
Move CLI plugins to "/usr/local/libexec" (as they're not packaging-managed)

tianon:


tianon:
I can't find any reason why I didn't use `/usr/local/libexec` to begin with. :shrug:

yosifkit (approved):

PR #424, opened 2023-05-10
Update to Alpine 3.18

jnoordsij:
See also https://alpinelinux.org/posts/Alpine-3.18.0-released.html.

### DIFF
--- a/20.10/cli/Dockerfile
+++ b/20.10/cli/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM alpine:3.17
+FROM alpine:3.18
 
 RUN apk add --no-cache \
 		ca-certificates \
--- a/23.0/cli/Dockerfile
+++ b/23.0/cli/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM alpine:3.17
+FROM alpine:3.18
 
 RUN apk add --no-cache \
 		ca-certificates \
--- a/24-rc/cli/Dockerfile
+++ b/24-rc/cli/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM alpine:3.17
+FROM alpine:3.18
 
 RUN apk add --no-cache \
 		ca-certificates \
--- a/Dockerfile-cli.template
+++ b/Dockerfile-cli.template
@@ -1,5 +1,5 @@
 {{ include "shared" -}}
-FROM alpine:3.17
+FROM alpine:3.18
 
 RUN apk add --no-cache \
 		ca-certificates \


andythedandy:
Nice....now millions of pipelines using still docker-compose won't run anymore.

jnoordsij:
> Nice....now millions of pipelines using still docker-compose won't run anymore.

I'm not sure what you mean by this, docker compose is still functional in the new image AFAICS:
```
$ docker container run --rm docker:cli /bin/sh -c docker-compose --help
Usage:  docker compose [OPTIONS] COMMAND

Docker Compose

Options:
      --ansi string                Control when to print ANSI control
                                   characters ("never"|"always"|"auto")
                                   (default "auto")
      --compatibility              Run compose in backward compatibility mode
      --env-file stringArray       Specify an alternate environment file.
  -f, --file stringArray           Compose configuration files
      --parallel int               Control max parallelism, -1 for
                                   unlimited (default -1)
      --profile stringArray        Specify a profile to enable
      --project-directory string   Specify an alternate working directory
                                   (default: the path of the, first
                                   specified, Compose file)
  -p, --project-name string        Project name

Commands:
  build       Build or rebuild services
  config      Parse, resolve and render compose file in canonical format
  cp          Copy files/folders between a service container and the local filesystem
  create      Creates containers for a service.
  down        Stop and remove containers, networks
  events      Receive real time events from containers.
  exec        Execute a command in a running container.
  images      List images used by the created containers
  kill        Force stop service containers.
  logs        View output from containers
  ls          List running compose projects
  pause       Pause services
  port        Print the public port for a port binding.
  ps          List containers
  pull        Pull service images
  push        Push service images
  restart     Restart service containers
  rm          Removes stopped service containers
  run         Run a one-off command on a service.
  start       Start services
  stop        Stop services
  top         Display the running processes
  unpause     Unpause services
  up          Create and start containers
  version     Show the Docker Compose version information

Run 'docker compose COMMAND --help' for more information on a command.
```

tianon:
Indeed, a little elaboration would be helpful :sweat_smile:

andythedandy:
Sorry. Yes. What you mean, is the "new" docker compose....this is indeed in there.

My CI pipelines were still using `docker-compose`, so they stopped working. The switch to `docker compose` was a no-brainer of course, but we have multiple services relying on a gitlab-ci template, that is using this command for creating an environment for integration testing.

Here you go for some context:
```
$ apk add --no-cache docker-compose
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/community/x86_64/APKINDEX.tar.gz
ERROR: unable to select packages:
  docker-compose (no such package):
    required by: world[docker-compose]
```

jnoordsij:
I understand the confusion now! When running `apk add --no-cache docker-compose`, you are actually installing the (old v1) `docker-compose` binary from the Alpine package registry; however, the `docker-compose` plugin is already pre-installed in this image, so installing it is unnecessary. 

The issue is this package is removed from Alpine as docker compose v1 is no longer supported. Note this is mostly unrelated to the `docker-compose` vs. `docker compose` syntax change, as the new v2 version still also support using `docker-compose` (see also my example above).

So my guess is you can remove the `apk add --no-cache docker-compose` from your workflow and everything should be working exactly as before, since the binary installed by `apk` was most likely ignored before. Alternatively, if you really need the v1 binary, you could just use `alpine:3.17` as base image and install the old package (see https://pkgs.alpinelinux.org/package/v3.17/community/x86_64/docker-compose); the new binary is now installable through the `docker-cli-compose` package (see https://pkgs.alpinelinux.org/package/v3.18/community/x86_64/docker-cli-compose).

Tl;dr: remove `apk add --no-cache docker-compose` and you should most likely be fine.

andythedandy:
All cool. As I said, I just removed the "-" from the lines in the gitlab-ci.yaml and removed the line, where the docker-compose apk was installed. So it's better then before, but some of us run in circles in panic, as we first need to find out what was wrong/what has changed ^^

tianon:
Just to thread the needle all the way back, https://github.com/docker-library/docker/pull/361 is where `docker-compose` was added to the image. :blush:

PR #430, opened 2023-06-01
Disable slowdown when TLS deliberately disabled

perlun:
Noticed this by coincidence when upgrading our local image to 23.0.6. The warning is fine in itself, but pushing the 15-second delay to all users setting `DOCKER_TLS_CERTDIR: ''` (perhaps for perfectly justifiable reasons) is a bit too harsh for my taste. :thinking: 

(Untested if this works with 24.0, so this should probably be verified before merging.)

### DIFF
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -127,10 +127,12 @@ if [ "$#" -eq 0 ] || [ "${1#-}" != "$1" ]; then
 			"$@"
 		DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS="${DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS:-} -p 0.0.0.0:2376:2376/tcp"
 	else
-		# TLS disabled (-e DOCKER_TLS_CERTDIR='') or missing certs
+		# TLS disabled (-e DOCKER_TLS_CERTDIR=''). # --tls=false silences the "Binding to IP address without --tlsverify is insecure" warning and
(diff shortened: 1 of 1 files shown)


perlun:
I get that point. :+1: I just think it's bad to push this delay onto the users, who might be unaware of it.

We've been using a Docker >= 19.03.12 version (where TLS is enabled by default) since June 2021, without fully realizing that this would cost us 15s for **every single CI run** using a Docker service in our local GitLab. :joy: I don't know for sure, but I wouldn't be very surprised if others are in the "same boat".

We ended up overriding it by setting `ENTRYPOINT ["dockerd-entrypoint.sh", "--tls=false"]` in our Docker image.

----

I'm open to alternative suggestions. Would you feel better about adding a note about this in the project `README.md` instead, which gets displayed at Docker Hub?

tianon:
See https://docs.docker.com/engine/deprecated/#unauthenticated-tcp-connections and https://docs.docker.com/go/api-security/ (using unauthenticated TCP is officially deprecated upstream, and as such is not something we will or even can continue to encourage or support).

perlun:
Interesting, thanks for the note @tianon. How is this planned (from Docker Inc) to work with Docker-in-Docker scenarios, do you know? :thinking: I mean, those are also perfectly valid "local" cases even though binding to `tcp://localhost` isn't enough for that use case. Enforcing TLS there will only add extra overhead/work for little benefit. Do you happen to know if this has been discussed upstream within the Moby project or so?

Issue #443, opened 2023-09-03
Host networking fails due to legacy iptables config


lukechilds:
If you want to run a second Docker daemon via dind with host networking access it doesn't currently work because almost all distros use the modern nftables backend for iptables and for some reason Alpine (which this dind image is based on) defaults to the legacy backend. If you run `docker run --privileged --net host docker:dind` that Docker instance will be using a totally different iptables backend than the host, the two will conflict, and containers run inside the dind instance with ports exposed will not be accessible on the host unless you manually update the host's iptables config.

You can see an example of this here:

```shell
# Host
ubuntu@host:~$ sudo iptables --version
iptables v1.8.7 (nf_tables)
ubuntu@host:~$ sudo iptables -S                                                                                                         
-P INPUT ACCEPT                                                                                                                         
-P FORWARD DROP                                                                                                                         
-P OUTPUT ACCEPT                                                                                                                        
-N DOCKER                                                                                                                               
-N DOCKER-ISOLATION-STAGE-1                                                                                                             
-N DOCKER-ISOLATION-STAGE-2                                                                                                             
-N DOCKER-USER                                                                                                                          
-A FORWARD -j DOCKER-USER                                                                                                               
-A FORWARD -j DOCKER-ISOLATION-STAGE-1                                                                                                  
-A FORWARD -o docker0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT                                                              
-A FORWARD -o docker0 -j DOCKER                                                                                                         
-A FORWARD -i docker0 ! -o docker0 -j ACCEPT                                                                                            
-A FORWARD -i docker0 -o docker0 -j ACCEPT                                                                                              
-A DOCKER-ISOLATION-STAGE-1 -i docker0 ! -o docker0 -j DOCKER-ISOLATION-STAGE-2                                                         
-A DOCKER-ISOLATION-STAGE-1 -j RETURN                                                                                                   
-A DOCKER-ISOLATION-STAGE-2 -o docker0 -j DROP                                                                                          
-A DOCKER-ISOLATION-STAGE-2 -j RETURN                                                                                                   
-A DOCKER-USER -j RETURN

# docker:dind
ubuntu@host:~$ sudo docker run --privileged --net host docker:dind iptables --version
iptables v1.8.9 (legacy)
ubuntu@host:~$ sudo docker run --privileged --net host docker:dind iptables -S
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
[...]

PR #444, opened 2023-09-03
Fix host networking

lukechilds:
Resolves https://github.com/docker-library/docker/issues/443

I think the iptables behaviour is a bug but the `DOCKER_ENSURE_BRIDGE` behaviour might be considered out of scope. It's pretty handy if you're using host networking though and it shouldn't affect anyone that doesn't have it set so it would be great to include it if people think it's appropriate.

### DIFF
--- a/23/dind/Dockerfile
+++ b/23/dind/Dockerfile
@@ -28,6 +28,13 @@ RUN set -eux; \
 		apk add --no-cache zfs; \
 	fi
 
+# Ensure we use nftables backend for iptables since legacy is the default on Alpine
+# which breaks when dind is used with host networking: https://github.com/docker-library/docker/issues/443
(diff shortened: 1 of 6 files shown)


Luisvelix (commented):
[Uploading CHANGELOG.txt…]()

Luisvelix:
[README-win32.txt](https://github.com/docker-library/docker/files/12643082/README-win32.txt)

tianon:
See https://github.com/docker-library/docker/pull/437#issuecomment-1845920765

tianon:
Closing in favor of https://github.com/docker-library/docker/pull/437 (`DOCKER_ENSURE_BRIDGE` is definitely out of scope, sorry!)

Issue #325, opened 2021-08-05
Missing os/arch if using specific version tag


jokay:
I would like to use the specific version tag but it lacks some os/arch *(linux/arm/v6, linux/arm/v7)*.

`:latest`

![image](https://user-images.githubusercontent.com/18613935/128308599-ab2cb846-3da1-40aa-8ddf-04bcdb71dd91.png)

`:20.10.8`

![image](https://user-images.githubusercontent.com/18613935/128308676-1abf0653-9c7b-4f43-b597-b53f63eb0610.png)

Would it be possible to have all os/arch from `:latest` in the specific version releases as well?

yosifkit:
That'd be great, but those architectures have been failing to build for quite some time (duplicate of https://github.com/docker-library/docker/issues/260).  It's likely that we'll just disable the `arm32*` image builds since `latest` is pulling in their last successful build from ~1 year ago (https://hub.docker.com/r/arm32v7/docker/tags?page=1&ordering=last_updated&name=latest). They can definitely come back once the problem is fixed upstream.

```console
+ dockerd --version
Bus error (core dumped)
```

jokay:
Thx for the information, will close the issue then :+1:

Issue #260, opened 2020-10-07
Docker 19.03.13 for arm32v7 build failing


kereis:
For arm32v7, the current version of Docker on Docker Hub is 19.03.08. Meanwhile, Docker version 19.03.13 is available (See [Jenkins](https://doi-janky.infosiftr.net/job/multiarch/job/arm32v7/job/docker/lastFailedBuild/)).

I think it shouldn't make a huge difference to either use 19.03.08 or 19.03.13, but it's a little bit inconsistent though and it could possibly mean that builds with any other newer version may fail.

tianon:
Indeed, the failure looks like the official binaries might be no good:

```
Step 6/12 : RUN set -eux; 		apkArch="$(apk --print-arch)"; 	case "$apkArch" in 		x86_64) dockerArch='x86_64' ;; 		armhf) dockerArch='armel' ;; 		armv7) dockerArch='armhf' ;; 		aarch64) dockerArch='aarch64' ;; 		*) echo >&2 "error: unsupported architecture ($apkArch)"; exit 1 ;;	esac; 		if ! wget -O docker.tgz "https://download.docker.com/linux/static/${DOCKER_CHANNEL}/${dockerArch}/docker-${DOCKER_VERSION}.tgz"; then 		echo >&2 "error: failed to download 'docker-${DOCKER_VERSION}' from '${DOCKER_CHANNEL}' for '${dockerArch}'"; 		exit 1; 	fi; 		tar --extract 		--file docker.tgz 		--strip-components 1 		--directory /usr/local/bin/ 	; 	rm docker.tgz; 		dockerd --version; 	docker --version
 ---> Running in 2003701aeec2
+ apk --print-arch
+ apkArch=armv7
+ dockerArch=armhf
+ wget -O docker.tgz https://download.docker.com/linux/static/stable/armhf/docker-19.03.13.tgz
Connecting to download.docker.com (13.225.65.117:443)
saving to 'docker.tgz'
docker.tgz           100% |********************************| 53.7M  0:00:00 ETA
'docker.tgz' saved
+ tar --extract --file docker.tgz --strip-components 1 --directory /usr/local/bin/
+ rm docker.tgz
+ dockerd --version
Illegal instruction (core dumped)
```

:disappointed:

djmaze:
[...]

PR #326, opened 2021-08-05
Drop arm32v6 and arm32v7 builds for now

yosifkit:
[arm32v6](https://doi-janky.infosiftr.net/job/multiarch/view/images/view/docker/job/arm32v6/job/docker/): still broken
```console
+ bashbrew build docker:20.10.8
...
+ dockerd --version
Bus error (core dumped)

+ bashbrew build docker:19.03.15
...
+ dockerd --version
Segmentation fault (core dumped)
```

[arm32v7](https://doi-janky.infosiftr.net/job/multiarch/view/images/view/docker/job/arm32v7/job/docker/): still broken
```console
+ Bus error docker:20.10.8
...
+ dockerd --version
Illegal instruction (core dumped)

+ bashbrew build docker:19.03.15
...
+ dockerd --version
Segmentation fault (core dumped)
```

ref https://github.com/docker-library/docker/issues/260
closes https://github.com/docker-library/docker/issues/325

cc @thaJeztah anything @tianon and I can do to help debug the upstream `docker` builds for `arm32*`?

PR #447, opened 2023-09-28
re-enable arm32v6 and arm32v7 builds

thaJeztah:
- fixes https://github.com/docker-library/docker/issues/260
- fixes https://github.com/docker-library/docker/issues/325
- updates https://github.com/docker-library/docker/pull/326
- relates to https://github.com/moby/moby/issues/31254#issuecomment-1738675671

Builds for these architectures were broken for some time, but this should be resolved now, so we can re-enable these (fingers crossed).

This reverts commit 461a3ec6ed346995c5949cc2d70c08b9d049caa0.

### DIFF
--- a/generate-stackbrew-library.sh
+++ b/generate-stackbrew-library.sh
@@ -81,9 +81,6 @@ versionArches() {
 				.[env.version].arches | to_entries[]
 				| select(.value.'"$selector"')
 				| .key
-				# all arm32 builds are broken:
-				# https://github.com/docker-library/docker/issues/260
-				| select(startswith("arm32") | not)
 			' versions.json | sort
 		) \
 		<(xargs -n1 <<<"$parentArches" | sort)


thaJeztah:
@tianon @yosifkit PTAL (not exactly sure how to verify this locally, so just push the PR)

thaJeztah:
CI looks happy, and I see some `armhf` / `armv6` strings in the logs (but not sure if it actually did those builds and tested them; it's possible other changes are needed to re-enable 🙈)

yosifkit (approved):
Successfully built `arm32v6` and `arm32v7` locally with qemu-user-static and also built `arm32v7` on an `arm64v8` aws server.

thaJeztah:
Thanks for checking!

tianon:
I tested on an arm32v7 box and they seem OK; my v6 box is less accessible, so it seems pretty reasonable IMO to roll the dice and assume it's fixed. :smile:

PR #450, opened 2023-10-03
Added inline SBOM for binaries downloaded outside package manager

LaurentGoderre:


Dockerfile-cli.template
```diff
@@ -1,5 +1,6 @@
 {{ include "shared" -}}
-FROM alpine:3.18
+{{ include "template-helper-functions" }}
```
yosifkit, line 2: Need the `-` to tell it to slurp up the whitespace (so that it doesn't generate a blank line in the final Dockerfiles). It could be combined with the other `include`, but is fine separate as well.

```suggestion
{{ include "template-helper-functions" -}}
```

Dockerfile-dind.template
```diff
@@ -1,4 +1,5 @@
 {{ include "shared" -}}
+{{ include "template-helper-functions" }}
```
yosifkit, line 2: And here.

```suggestion
{{ include "template-helper-functions" -}}
```

Dockerfile-dind.template
```diff
@@ -53,6 +54,33 @@ RUN set -eux; \
...
+			version: .version,
+			params: {
+				os_name: "alpine",
+				os_version: .alpine
+			},
+			licenses: [
+				"Apache-2.0"
+			]
+		} | sbom | tostring
+	}}' > /usr/local/bin/docker-compose.spdx.json; \
```
yosifkit, line 82: The `dind` image shouldn't need a new `docker-compose` spdx since it is `FROM` the cli image that should already add that.  As for the `docker` one, that should maybe be named `dockerd` since the `cli` image already added a `docker` for the Docker cli binary.

Dockerfile-cli.template
```diff
@@ -65,6 +80,20 @@ RUN set -eux; \
...
+			version: .version,
+			params: {
+				os_name: "alpine",
+				os_version: "3.18"
+			},
+			licenses: [
+				"Apache-2.0"
+			]
+		} | sbom | tostring
+	}}' > /usr/local/bin/docker-{{ $key }}.spdx.json ; \
```
yosifkit, line 96: This is only getting added for `docker-compose`, but the other binary of the loop is getting missed (`buildx`). Probably just needs to be moved above the `if $key == "compose"` line?

Dockerfile-dind-rootless.template
```diff
@@ -35,6 +36,37 @@ RUN set -eux; \
...
+			licenses: [
+				"Apache-2.0"
+			]
+		} | sbom | tostring
+	}}' > /usr/local/bin/docker-compose.spdx.json; \
+	\
+	dockerd --version; \
+	containerd --version; \
+	ctr --version; \
+	runc --version; \
```
yosifkit, line 69: These binaries all come from the `dind` image that this is from; they are already tested when downloaded rather than in a dependent image to make sure a broken binary doesn't get committed to an image.

PR #461, opened 2023-12-07
Update to Alpine 3.19

jnoordsij:
This updates Alpine to the latest stable version: 3.19.

See also https://alpinelinux.org/posts/Alpine-3.19.0-released.html.

*Edit: This is currently blocked as it requires additional upstream support for `nftables`; see also [comment below](https://github.com/docker-library/docker/pull/461#issuecomment-1845922045) and #437.*

Closes: https://github.com/docker-library/docker/pull/437

### DIFF
--- a/24/cli/Dockerfile
+++ b/24/cli/Dockerfile
@@ -4,7 +4,7 @@
 # PLEASE DO NOT EDIT IT DIRECTLY.
 #
 
-FROM alpine:3.18
+FROM alpine:3.19
 
 RUN apk add --no-cache \
 		ca-certificates \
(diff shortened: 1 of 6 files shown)


tianon:
Ah, https://github.com/docker-library/docker/pull/437 by itself is not correct/complete, but it _is_ a great thing to surface and a big problem for this image (surfacing a larger problem with iptables vs nftables, especially as it relates to network namespaces / Docker-in-Docker). :disappointed:

jnoordsij:
That seems like a rather large blocking issue then. I'll mark this as draft in the meantime, so anyone interested can still subscribe and track progress..

tianon (approved):
perfect IMO! :heart:

PR #462, opened 2023-12-13
Add a "docker" group

### LINKED ISSUE #459: (re)consider adding `docker` group to docker-in-docker (dind) image
There's two older tickets related to this;

- https://github.com/docker-library/docker/issues/55
- https://github.com/docker-library/docker/issues/382

The docker daemon by default tries to set permissions of the unix-socket to be owned by the `docker` group. This group is used to allow accessing the socket as a non-root user, as described in the ["Manage Docker as a non-root user"](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user) section in the documentation.

Currently, the `docker` images do not contain a `docker` group, which results in a warning being printed during startup of the daemon;

```bash
docker run -it --rm --privileged -v docker:/var/lib/docker docker:dind sh
/ # dockerd
INFO[2023-11-15T11:24:34.632589125Z] Starting up
WARN[2023-11-15T11:24:34.634358166Z] could not change group /var/run/docker.sock to docker: group docker not found
INFO[2023-11-15T11:24:34.635752416Z] containerd not running, starting managed containerd
```

This warning originates from the daemon, which attempts to use the default (`docker`) group as owner for the socket https://github.com/moby/moby/blob/93fffa299c2cb23055bb54c380a02d53c9a7d525/daemon/listeners/listeners_linux.go#L35-L49

```go
case "unix":
	gid, err := lookupGID(socketGroup)
	if err != nil {
		if socketGroup != "" {
			if socketGroup != defaultSocketGroup {
				return nil, err
			}
			log.G(context.TODO()).Warnf("could not change group %s to %s: %v", addr, defaultSocketGroup, err)
		}
		gid = os.Getgid()
	}
	l, err := sockets.NewUnixSocket(addr, gid)
	if err != nil {
		return nil, errors.Wrapf(err, "can't create unix socket %s", addr)
	}
```

> note: we should probably update the [Manage Docker as a non-root user](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user) documentation a bit though as the instructions _assume_ the `docker` group has to be created (we should mention how to check if it already exists). 

I went looking what we do in other situations, i.e., when installing from our `.deb` and `.rpm` packages, and for those, we are creating the `docker` group as part of the installation; https://github.com/docker/docker-ce-packaging/blob/3479bfbc96ec430a808b8d22a503c12773463a66/deb/common/docker-ce.postinst#L5-L11

```bash
configure)
	if [ -z "$2" ]; then
		if ! getent group docker > /dev/null; then
			groupadd --system docker
		fi
	fi
	;;
```

And, at runtime, that group is used as owner for the socket, as well as log-files produced;

### With systemd

https://github.com/moby/moby/blob/93fffa299c2cb23055bb54c380a02d53c9a7d525/contrib/init/systemd/docker.socket#L4-L10

```unit file (systemd)
[Socket]
# If /var/run is not implemented as a symlink to /run, you may need to
# specify ListenStream=/var/run/docker.sock instead.
ListenStream=/run/docker.sock
SocketMode=0660
SocketUser=root
SocketGroup=docker
```

### With openrc

https://github.com/moby/moby/blob/93fffa299c2cb23055bb54c380a02d53c9a7d525/contrib/init/openrc/docker.initd#L24-L26

```bash
start_pre() {
	checkpath -f -m 0644 -o root:docker "$DOCKER_LOGFILE"
}
```

### sysvinit

https://github.com/moby/moby/blob/93fffa299c2cb23055bb54c380a02d53c9a7d525/contrib/init/sysvinit-debian/docker#L59

```bash
chgrp docker "$DOCKER_LOGFILE"
```

## Consider adding a `docker` group

While users should still "think twice" before granting access to the unix-socket (as it is [equivalent of `root`](https://docs.docker.com/engine/security/#docker-daemon-attack-surface)), and consider rootless instead if their situation allows it, I think it would make sense to add a `docker` group to the image, so that we're consistent with other ways the docker daemon is distributed / packaged.

Perhaps there's reasons why we didn't choose to though, but if that's the case (and we decide _not to_), we can use this ticket to document those reasons (and possibly extend the documentation on Docker Hub).

Alternatively, we could update the entrypoint script to explicitly set `root` as group with the `-G` / `--group` flag; doing so should suppress the warning.

```
  -G, --group string                            Group for the unix socket (default "docker")
```

```bash
docker run -it --rm --privileged -v docker:/var/lib/docker docker:dind sh
/ # dockerd --group=root
INFO[2023-11-15T11:24:03.549368180Z] Starting up
INFO[2023-11-15T11:24:03.556428722Z] containerd not running, starting managed containerd
```

tianon:
(with GID 2375)

Closes #459

### DIFF
--- a/24/cli/Dockerfile
+++ b/24/cli/Dockerfile
@@ -17,6 +17,10 @@ RUN apk add --no-cache \
 # - docker run --rm debian:stretch grep '^hosts:' /etc/nsswitch.conf
 RUN [ -e /etc/nsswitch.conf ] && grep '^hosts: files dns' /etc/nsswitch.conf
 
+# pre-add a "docker" group for socket usage
+RUN set -eux; \
+	addgroup -g 2375 -S docker
+
 ENV DOCKER_VERSION 24.0.7
 
 RUN set -eux; \
--- a/25-rc/cli/Dockerfile
+++ b/25-rc/cli/Dockerfile
@@ -17,6 +17,10 @@ RUN apk add --no-cache \
 # - docker run --rm debian:stretch grep '^hosts:' /etc/nsswitch.conf
 RUN [ -e /etc/nsswitch.conf ] && grep '^hosts: files dns' /etc/nsswitch.conf
 
+# pre-add a "docker" group for socket usage
+RUN set -eux; \
+	addgroup -g 2375 -S docker
+
 ENV DOCKER_VERSION 25.0.0-beta.2
 
 RUN set -eux; \
--- a/Dockerfile-cli.template
+++ b/Dockerfile-cli.template
@@ -12,6 +12,10 @@ RUN apk add --no-cache \
 # - docker run --rm debian:stretch grep '^hosts:' /etc/nsswitch.conf
 RUN [ -e /etc/nsswitch.conf ] && grep '^hosts: files dns' /etc/nsswitch.conf
 
+# pre-add a "docker" group for socket usage
+RUN set -eux; \
+	addgroup -g 2375 -S docker
+
 ENV DOCKER_VERSION {{ .version }}
 
 RUN set -eux; \


yosifkit:
Maybe the rootless user needs to be in the group: https://github.com/docker-library/docker/issues/459#issuecomment-1854926581

yosifkit (approved):

tianon:
Copying https://github.com/docker-library/official-images/pull/15943#issuecomment-1866962674 here for visibility:

> Ok, I've applied https://github.com/docker-library/docker/commit/3a0424834732f20142291cdd1d392e6d860af2ad for now so we can get this in.  Since you're planning at least one RC before GA in (hopefully) Jan, I'll leave that as-is for now, and will plan to revert that revert after the holiday season.

Issue #464, opened 2023-12-15
IPTables Change, can't run Gitlab CI CD anymore


intsanerarity:
Hey,

with your newest Version i get this error from my CI/CD Pipeline 
"error during connect: Post "http://docker:2375/v1.24/auth": dial tcp: lookup docker on 169.254.169.254:53: no such host"

i set dind fix on 23.0.6 and everything is working fine - any suggestions?

best regards

steven

manneorama:
Same here but on Github using `action-runner-controller` in GKE with `docker:dind` as a sidecar. Worked around it by pinning `docker:dind` to the previous version in the runner controller Helm chart.

stevexuereb:
@manneorama  This is a duplicate of https://github.com/docker-library/docker/issues/463, for GitLab.com specifics you can check https://gitlab.com/gitlab-com/gl-infra/production/-/issues/17283

tianon:
Duplicate of https://github.com/docker-library/docker/issues/463 :+1: :heart:

PR #465, opened 2023-12-15
Switch over to xtables-legacy when nf_tables module isn't available

### LINKED ISSUE #463: dockerd fails to start - RULE_APPEND failed (No such file or directory): rule in chain DOCKER-ISOLATION-STAGE-1
After the merge of https://github.com/docker-library/docker/pull/461, the `docker` containers we run as part of our CI jobs stopped working.
The container fails to start with the following error:
```
failed to start daemon: Error initializing network controller: error obtaining controller instance: unable to add return rule in DOCKER-ISOLATION-STAGE-1 chain:  (iptables failed: iptables --wait -A DOCKER-ISOLATION-STAGE-1 -j RETURN: iptables v1.8.10 (nf_tables):  RULE_APPEND failed (No such file or directory): rule in chain DOCKER-ISOLATION-STAGE-1
```

The affected image is `docker.io/library/docker@sha256:ae63bb7c7d3ae23884a2c5d206939640279f6d15730618192b58662a0619f182`, while `docker.io/library/docker@sha256:c90e58d30700470fc59bdaaf802340fd25c1db628756d7bf74e100c566ba9589` works fine. Both images are tagged as `24.0.7-dind`

The environment is GKE 1.27 with Container-Optimized OS.

**Workaround**
Use `docker:24.0.7-dind-alpine3.18`, as it points at the previous version of the image that was overwritten

akerouanton:
* Closes #461
* Closes #463
* Related to https://gitlab.com/gitlab-com/gl-infra/production/-/issues/17283

PR #461 updated Alpine to 3.19 and made a change to load the nf_tables kernel module if needed. However, as demonstrated by #463 and #464 this might break when the host system doesn't have the nf_tables module available. In that case, we should still try to load the ip_tables module and symlink /sbin/iptables to xtables-legacy-multi.

### DIFF
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -143,12 +143,22 @@ if [ "$1" = 'dockerd' ]; then
 	# XXX inject "docker-init" (tini) as pid1 to workaround https://github.com/docker-library/docker/issues/318 (zombie container-shim processes)
 	set -- docker-init -- "$@"
 
+	use_xtables_legacy=false
 	if ! iptables -nL > /dev/null 2>&1; then
 		# if iptables fails to run, chances are high the necessary kernel modules aren't loaded (perhaps the host is using nftables with the translating "iptables" wrappers, for example)
 		# https://github.com/docker-library/docker/issues/350
 		# https://github.com/moby/moby/issues/26824
 		# https://github.com/docker-library/docker/pull/437#issuecomment-1854900620
-		modprobe nf_tables || :
+		if ! modprobe nf_tables; then
+			modprobe ip_tables || :
+			use_xtables_legacy=true
+		fi
+	fi
+	if [ "$use_xtables_legacy" = "true" ]; then
+		ln -fs /sbin/iptables-legacy /sbin/iptables
+		# iptables-restore and iptables-save aren't used by dockerd currently, but let's not ship a half broken image.
+		ln -fs /sbin/iptables-legacy-restore /sbin/iptables-restore
+		ln -fs /sbin/iptables-legacy-save /sbin/iptables-save
 	fi
 
 	uid="$(id -u)"
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -143,22 +143,16 @@ if [ "$1" = 'dockerd' ]; then
 	# XXX inject "docker-init" (tini) as pid1 to workaround https://github.com/docker-library/docker/issues/318 (zombie container-shim processes)
 	set -- docker-init -- "$@"
 
-	use_xtables_legacy=false
 	if ! iptables -nL > /dev/null 2>&1; then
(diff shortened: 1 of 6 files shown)


dockerd-entrypoint.sh
```diff
@@ -143,12 +143,22 @@ if [ "$1" = 'dockerd' ]; then
...
 		# https://github.com/docker-library/docker/pull/437#issuecomment-1854900620
-		modprobe nf_tables || :
+		if ! modprobe nf_tables; then
+			modprobe ip_tables || :
+			use_xtables_legacy=true
+		fi
+	fi
+	if [ "$use_xtables_legacy" = "true" ]; then
+		ln -fs /sbin/iptables-legacy /sbin/iptables
+		# iptables-restore and iptables-save aren't used by dockerd currently, but let's not ship a half broken image.
```
robmry, line 159: Looks like the alpine change [sets up 'ip6tables' links](https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/47102/diffs#5e2153cf80a0441cf9b7bff6df4cf7eca9aaf222_60_76) as well as the 'iptables' ones. Perhaps best to modify those too?

tianon:
Nice, thanks for working on this! :heart:

Modifying `/sbin/xxx` at runtime is definitely making me really nervous (both overwriting files installed by a package _and_ modifying `/sbin/` in general), so I think I'd prefer if we create a directory of symlinks in the `Dockerfile` that the entrypoint can then use `PATH` tricks to activate instead.  Given we're already at the end of your day (and the severity of the breaking change), I hope you don't mind me taking over instead of describing in more detail and requesting changes? :innocent:

dockerd-entrypoint.sh
```diff
@@ -144,11 +144,15 @@ if [ "$1" = 'dockerd' ]; then
...
-		# if iptables fails to run, chances are high the necessary kernel modules aren't loaded (perhaps the host is using nftables with the translating "iptables" wrappers, for example)
+		# if iptables fails to run, chances are high the necessary kernel modules aren't loaded (perhaps the host is using xtables, for example)
 		# https://github.com/docker-library/docker/issues/350
 		# https://github.com/moby/moby/issues/26824
 		# https://github.com/docker-library/docker/pull/437#issuecomment-1854900620
-		modprobe nf_tables || :
+		if ! modprobe nf_tables; then
+			modprobe ip_tables || :
+			# see https://github.com/docker-library/docker/issues/463 (and the dind Dockerfile where this directory is set up)
+			export PATH="/usr/local/sbin/.iptables-legacy:$PATH"
```
tianon, line 154: A nice side benefit of this method is that users having trouble with it misfiring (somehow) is that they can just hide this directory somehow like bind-mounting `/dev/null` on top of it, and the container will happily continue using the default Alpine-provided `iptables` symlink. :+1:

tianon:
```console
/ # ls /usr/local/
bin    lib    share
```

grumble grumble grumble

yosifkit (approved):

tianon:
Follow-up in https://github.com/docker-library/docker/pull/468 :eyes:

PR #483, opened 2024-02-23
Add `git` to `dind` variants (and make `git` an alias of `dind`, for 24.x and 25.x)

### LINKED ISSUE #482: Include `git` or otherwise address "current commit information was not captured by the build: git was not found in the system: ..."
When I use the `docker` image as a build step in Google Cloud Build to perform buildx builds, I get the following warning:

    WARNING: current commit information was not captured by the build: git was not found in the system: exec: "git": executable file not found in $PATH

I presume this has to do with provenance being enabled by default per https://docs.docker.com/build/attestations/slsa-provenance/#mode

I would like to be able to use the `docker` image to run `docker buildx ...` and have the provenance information captured as designed by the new default. I know about the option to suppress the warning by using `BUILDX_NO_DEFAULT_ATTESTATIONS=1` to disable the feature entirely, but I would prefer to get it working instead.

As an end user I have no preference between what I presume are the two possible solutions:

1. Modify the provenance feature to extract the required information without running `git` as a child process.
2. Add `git` to the `docker` image.

(I'm opening this issue here because I presume the second option is much more reasonable to implement.)

tianon:
See https://github.com/docker-library/docker/issues/482#issuecomment-1955609423
Closes #482

(with [apologies](https://github.com/docker-library/docker/pull/295) to @thaJeztah :heart:)

This also does some minor cleanup of our installation of other dependencies.

### DIFF
--- a/Dockerfile-dind.template
+++ b/Dockerfile-dind.template
@@ -1,29 +1,23 @@
 {{ include "shared" -}}
 FROM docker:{{ env.version }}-cli
 
-# https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
+# https://github.com/moby/moby/blob/0eecd59153c03ced5f5ddd79cc98f29e4d86daec/project/PACKAGERS.md#runtime-dependencies
+# https://github.com/docker/docker-ce-packaging/blob/963aa02666035d4e268f33c63d7868d6cdd1d34c/deb/common/control#L28-L41
 RUN set -eux; \
 	apk add --no-cache \
 		btrfs-progs \
 		e2fsprogs \
 		e2fsprogs-extra \
+		git \
 		ip6tables \
 		iptables \
 		openssl \
+		pigz \
 		shadow-uidmap \
 		xfsprogs \
 		xz \
-# pigz: https://github.com/moby/moby/pull/35697 (faster gzip implementation)
-		pigz \
-	; \
-# only install zfs if it's available for the current architecture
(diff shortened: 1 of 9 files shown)


yosifkit (approved):

Dockerfile-dind.template
```diff
@@ -1,29 +1,23 @@
...
-	; \
-# only install zfs if it's available for the current architecture
-# https://git.alpinelinux.org/cgit/aports/tree/main/zfs/APKBUILD?h=3.6-stable#n9 ("all !armhf !ppc64le" as of 2017-11-01)
-# "apk info XYZ" exits with a zero exit code but no output when the package exists but not for this arch
-	if zfs="$(apk info --no-cache --quiet zfs)" && [ -n "$zfs" ]; then \
-		apk add --no-cache zfs; \
-	fi
-
-# TODO aufs-tools
+		zfs \
```
thaJeztah, line 19: Actually never really realised we're including `zfs` and `btrfs` in these images. With our "best effort" move towards overlayfs on all underlying filesystems, I'm also wondering how common it is to run DIND with zfs or btrfs.

thaJeztah (commented):
"LGTM", thanks!

Dockerfile-dind.template
tianon: Yeah, as I changed this I was thinking about how `zfs` is probably going away entirely. :joy:

I think things break if we don't have these and users _are_ using them though, so :shrug:

Both of these together are still smaller than Git :joy:

PR #488, opened 2024-04-02
Update "versions" curl invocation to be more defensive

tianon:
This allows the script to differentiate 404 from 503 so that we are less likely to accidentally commit downgrades (like https://github.com/docker-library/docker/commit/9b7730d06c32e4e5521e8a3061bd200a77f2132a) for temporary failures.

Example output if I change the `case` to not match 200 anymore:

```console
$ ./versions.sh
error: unexpected status code 200 while fetching: -H Accept: application/json https://github.com/docker/docker/commits/master/hack/dind.atom
Terminated
$ 
```

### DIFF
--- a/versions.sh
+++ b/versions.sh
@@ -23,8 +23,22 @@ else
 fi
 versions=( "${versions[@]%/}" )
 
+scriptPid="$$" # so we can kill the script even from a subshell
+_curl() {
+	exec 42>&1
+	local code
+	code="$(curl --silent --location -o /dev/fd/42 --write-out '%{http_code}' "$@")"
+	case "$code" in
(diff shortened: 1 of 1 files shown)


versions.sh
```diff
@@ -23,8 +23,22 @@ else
 fi
 versions=( "${versions[@]%/}" )
 
+scriptPid="$$" # so we can kill the script even from a subshell
+_curl() {
+	exec 42>&1
```
whalelines, line 28: Cute

whalelines (approved):
LGTM

yosifkit (approved):

PR #501, opened 2024-06-19
Also load `ip6_tables` when trying to load `ip_tables`

tianon:
This isn't used/necessary in the default configuration until Docker 27+, but it was optional behavior before that, so it's prudent for us to also try loading it any time we know we're not using nftables / have explicitly requested "legacy" `iptables`.

See also:
- https://github.com/moby/moby/pull/47747
- https://github.com/moby/moby/issues/47895
- https://github.com/moby/moby/pull/47918
- https://github.com/moby/moby/pull/47960

cc/fyi @robmry @akerouanton :heart:

### DIFF
--- a/24/dind/dockerd-entrypoint.sh
+++ b/24/dind/dockerd-entrypoint.sh
@@ -149,6 +149,7 @@ if [ "$1" = 'dockerd' ]; then
 		iptablesLegacy="$DOCKER_IPTABLES_LEGACY"
 		if [ -n "$iptablesLegacy" ]; then
 			modprobe ip_tables || :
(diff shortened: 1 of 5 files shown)


tianon:
I thought we'd already been doing this, but apparently I was misremembering, probably because we _do_ check the `/proc/net/ip6_tables_names` file for clues to whether we should be using legacy xtables. :sweat_smile:

whalelines (approved):
LGTM

akerouanton (commented):
LGTM

yosifkit (approved):

PR #504, opened 2024-07-01
Add `git` to CLI variants

tianon:
- `docker build .` / `docker buildx build .` might shell out to git to ask about the context directory's `git status`
- `DOCKER_BUILDKIT=0 docker build https://example/foo.git` will run git in the client context
- `docker buildx build https://example/foo.git` will run git in the daemon context

https://github.com/docker-library/docker/issues/482#issuecomment-2197116408

### DIFF
--- a/24/cli/Dockerfile
+++ b/24/cli/Dockerfile
@@ -9,7 +9,9 @@ FROM alpine:3.20
 RUN apk add --no-cache \
 		ca-certificates \
 # DOCKER_HOST=ssh://... -- https://github.com/docker/cli/pull/1014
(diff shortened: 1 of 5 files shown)


tianon:
For now, I've just added `git` to the CLI variants without removing it from the DinD variants because the later `apk add` should be a no-op, but I'm open to removing it and relying on the `FROM` relationship

tao12345666333 (commented):
LGTM

I think it's good that DIND keeps git.

yosifkit (approved):

PR #527, opened 2025-02-20
Update to 28.0.0 (GA)

tianon:
Also, drop 27.x (now EOL)


tianon:
I forgot the bot would pick up 28 successfully (https://github.com/docker-library/docker/commit/35c016d0c0f0be4dc523bc1108d7fa2681db8306), so now this is just dropping 27. :sweat_smile:

tianon:
Windows failures are https://github.com/docker-library/official-images/pull/18435#issuecomment-2654277193:

> The problem is that the `windows-2025` runner is too old to run the `mcr.microsoft.com/windows/servercore:ltsc2025` image released this week (https://github.com/actions/runner-images/issues/11582).

PR #536, opened 2025-05-19
Add additional extensions to generated ssl certs

jnoordsij:
This MR adds some additional extensions to the self-generated CA and server certificates the dind image may employ to set up TLS connections. This ensures the certificates (better) match RFCs specifying these extensions as required (see https://datatracker.ietf.org/doc/html/rfc3280 and https://datatracker.ietf.org/doc/html/rfc5280). This is particularly relevant for clients enforcing checks on this while connecting.

## Source of the problem

When using my combination of:
- GitLab CI 
- Docker dind image as service in CI job
- Python Docker image employing the [Docker SDK for Python](https://docker-py.readthedocs.io/en/stable/) to connect to the Docker daemon

I was presented with errors among the following lines:
> docker.errors.DockerException: Error while fetching server API version: HTTPSConnectionPool(host='docker', port=2376): Max retries exceeded with url: /version (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: CA cert does not include key usage extension (_ssl.c:1028)')))

This is due to a [recent change](https://github.com/urllib3/urllib3/issues/3571) in the Python `urllib3` package used by the Python Docker SDK, which follows the [stricter checks Python introduced with >=3.13](https://github.com/python/cpython/pull/112389). 

## MWE + proof of fix

I've created https://gitlab.com/jnoordsij/test-docker-ci/-/tree/e5289dcdc5a736663d66ea4199f95f5c94d73fec as basic example of attempting to use the Python Docker SDK with the dind image that fails; see https://gitlab.com/jnoordsij/test-docker-ci/-/jobs/10076832026.

After building a patched dind image (https://gitlab.com/jnoordsij/test-docker-ci/-/commit/0c6a1ac70fd6d6f26e29af8d5d797b3d70d41967) and using that in the CI job, everything seems to work fine
(see also https://gitlab.com/jnoordsij/test-docker-ci/-/pipelines/1824971402).

## Further notes

I think technically speaking only the second commit, updating the CA cert, is required to fix this particular issue. However when looking into this, I figured adding the server cert extensions is probably good practice, especially with this being recommended in official documentation.

### DIFF
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -53,6 +53,7 @@ _tls_generate_certs() {
 			-subj '/CN=docker:dind server'
 		cat > "$dir/server/openssl.cnf" <<-EOF
 			[ x509_exts ]
+			extendedKeyUsage = serverAuth
 			subjectAltName = $(_tls_san)
 		EOF
 		openssl x509 -req \
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -41,7 +41,10 @@ _tls_generate_certs() {
 		_tls_ensure_private "$dir/ca/key.pem"
 		openssl req -new -key "$dir/ca/key.pem" \
 			-out "$dir/ca/cert.pem" \
-			-subj '/CN=docker:dind CA' -x509 -days "$certValidDays"
+			-subj '/CN=docker:dind CA' \
+			-x509 \
+			-days "$certValidDays" \
+			-addext keyUsage=critical,digitalSignature,keyCertSign
 	fi
 
 	if [ -s "$dir/ca/key.pem" ]; then
--- a/28/dind/dockerd-entrypoint.sh
+++ b/28/dind/dockerd-entrypoint.sh
@@ -53,6 +53,7 @@ _tls_generate_certs() {
 			-subj '/CN=docker:dind server'
 		cat > "$dir/server/openssl.cnf" <<-EOF
 			[ x509_exts ]
+			extendedKeyUsage = serverAuth
 			subjectAltName = $(_tls_san)
 		EOF
 		openssl x509 -req \
--- a/28/dind/dockerd-entrypoint.sh
+++ b/28/dind/dockerd-entrypoint.sh
@@ -41,7 +41,10 @@ _tls_generate_certs() {
 		_tls_ensure_private "$dir/ca/key.pem"
 		openssl req -new -key "$dir/ca/key.pem" \
 			-out "$dir/ca/cert.pem" \
-			-subj '/CN=docker:dind CA' -x509 -days "$certValidDays"
+			-subj '/CN=docker:dind CA' \
+			-x509 \
+			-days "$certValidDays" \
+			-addext keyUsage=critical,digitalSignature,keyCertSign
 	fi
 
 	if [ -s "$dir/ca/key.pem" ]; then


tianon (approved):
Oof, thank you :heart:

This is 100% neither your fault nor your problem, but even as the original author of most of this code, I hate it - I hate that we have to have it, I hate how complicated it is, and I especially hate that it's so fiddly to get "right"

I would _love_ to deprecate it all, tell everyone to use the Unix socket instead, and call it a day, but that's not terribly realistic. :disappointed:

dockerd-entrypoint.sh
```diff
@@ -41,7 +41,10 @@ _tls_generate_certs() {
 		_tls_ensure_private "$dir/ca/key.pem"
 		openssl req -new -key "$dir/ca/key.pem" \
 			-out "$dir/ca/cert.pem" \
-			-subj '/CN=docker:dind CA' -x509 -days "$certValidDays"
+			-subj '/CN=docker:dind CA' \
+			-x509 \
+			-days "$certValidDays" \
+			-addext keyUsage=critical,digitalSignature,keyCertSign
```
tianon, line 47: All the OpenSSL docs say about `critical` here is this:

> If critical is present then the extension will be marked as critical.

Do you have a reference for why this is necessary? :thinking:

The docs also suggest this can be more whitespace-y, so if we need to touch this again I think I might try something like this:

```suggestion
			-addext 'keyUsage = critical, digitalSignature, keyCertSign'
```

But we should probably add a new test or update our existing test to make sure this stays working before we touch it again (since you've done good validation on this change as-is). :heart:

jnoordsij:
> Oof, thank you ❤️
> 
> This is 100% neither your fault nor your problem, but even as the original author of most of this code, I hate it - I hate that we have to have it, I hate how complicated it is, and I especially hate that it's so fiddly to get "right"
> 
> I would _love_ to deprecate it all, tell everyone to use the Unix socket instead, and call it a day, but that's not terribly realistic. 😞

Yeah I understand your feelings on this, it feels a bit too over-complicated to have this here. Deprecating it will probably be quite a hard thing to achieve. However, at least it does what it have to do right now pretty good, and I think generally it should be more or less a stable approach apart from cases like this enforcing a bit more strictness.

For changing towards sockets, just looking at GitLab for example, they do list [using a socket](https://docs.gitlab.com/ci/docker/using_docker_build/#use-a-unix-socket-on-a-shared-volume-between-docker-in-docker-and-build-container) as an option in their docs, but not as the preferred one (section starts with "You should use Docker-in-Docker with TLS enabled, which is supported by [GitLab.com instance runners](https://docs.gitlab.com/ci/runners/).") and moreover one that thus does not seem to be supported by their shared instance runners. 

Although I could not find current shared runner configuration in [their current docs](https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/ci/runners/hosted_runners/linux.md?ref_type=heads), an older version [only had a mount bind configured for the TLS based setup](https://gitlab.com/gitlab-org/gitlab/-/blob/cc807b373bd05a19654cfdb2f145bf339773a2bd/doc/ci/runners/saas/linux_saas_runner.md#configtoml). 

So before any of this could happen that would probably require quite a bit of effort in getting them to migrate, and then there's probably still lots of other usecases for the TLS approach out there 😓

dockerd-entrypoint.sh
jnoordsij: From the [RFC section 4.2.1.3 Key Usage](https://datatracker.ietf.org/doc/html/rfc3280#section-4.2.1.3):

> This extension MUST appear in certificates that contain public keys
   that are used to validate digital signatures on other public key
   certificates or CRLs.  When this extension appears, it SHOULD be
   marked critical.

PR #547, opened 2025-07-17
Use UID number rather than username in Dockerfile

candrews:
Systems configured to disallow running images as root aren't able to run images that use user name string values for the USER because they can't validate that a named user isn't root. To allow this image to run on such systems, use the uid of the user as the value for USER instead of the username.

See: https://github.com/kubernetes/kubernetes/pull/56503


tianon:
Closing as a duplicate of https://github.com/docker-library/docker/pull/442 (as you've noted over there).

PR #552, opened 2025-09-17
feat: use `util-linux` to replace `busybox`

laushunyu:
resolve https://github.com/docker-library/busybox/issues/230

### DIFF
--- a/28/cli/Dockerfile
+++ b/28/cli/Dockerfile
@@ -11,7 +11,9 @@ RUN apk add --no-cache \
 # DOCKER_HOST=ssh://... -- https://github.com/docker/cli/pull/1014
 		openssh-client \
 # https://github.com/docker-library/docker/issues/482#issuecomment-2197116408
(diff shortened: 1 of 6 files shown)


tianon:
I'm glad you found a solution to your problem!  However, I'm not sure I understand why we should apply it to all users of the `docker:dind` image (as opposed to you applying this locally in a custom image or deployment)?

Is there some functionality of `docker` or `dockerd` that's shelling out to `mount` and thus needs this to be functional for the edge case you ran into?

laushunyu:
I use docker:dind as gitlab staging environment, some developer's staging mount a specified cephfs to local dir and then map it to container using dind....

Currently, I’ve built another image based on `docker:dind` adding `mount` pkg, as you suggested.

Maybe..... I should open a issue of alpine to add the dependency `mount` to `ceph-fuse` ?

Anyway, thanks a lot! 😄

PR #554, opened 2025-10-06
Master

Strikkepinnene:


yosifkit:
1. We have no plans to use LLM tools like copilot.
2. There is not any python code that would require a test config for vscode (not that we would commit vscode config to the repo anyway)

PR #568, opened 2025-12-16
Add environment variable to control host loopback (`DOCKERD_ROOTLESS_ROOTLESSKIT_DISABLE_HOST_LOOPBACK`)

MadLittleMods:
Add environment variable to control host loopback (`DOCKERD_ROOTLESS_ROOTLESSKIT_DISABLE_HOST_LOOPBACK`)

Based on [`moby/moby` -> `contrib/dockerd-rootless.sh#L24-L25`](https://github.com/moby/moby/blob/8316b79e045e2b762373f0cdd87e4b1dbf22ea10/contrib/dockerd-rootless.sh#L24-L25) (added in https://github.com/moby/moby/pull/47352).

---

This kind of workaround is mentioned in various places like https://stackoverflow.com/questions/72500740/how-to-access-localhost-on-rootless-docker.

Our actual use case is setting up some self-hosted GitHub runners with a rootless Docker-in-Docker setup which works fine except that nested containers created within the `runner` container are unable to contact `host.docker.internal` (`Connection refused`).  (using [`actions/actions-runner-controller`](https://github.com/actions/actions-runner-controller/blob/3fd1048576883d9660fe02cec2a1556cb1387d17/charts/gha-runner-scale-set/values.yaml) ([docs](https://docs.github.com/en/actions/tutorials/use-actions-runner-controller/quickstart))). With this change, we plan to map `host.docker.internal:10.0.2.2` and does seem to work.

### DIFF
--- a/dockerd-entrypoint.sh
+++ b/dockerd-entrypoint.sh
@@ -1,4 +1,10 @@
 #!/bin/sh
+#
+# Recognized environment variables:
+# * DOCKERD_ROOTLESS_ROOTLESSKIT_NET=(slirp4netns|vpnkit): the rootlesskit network driver. Defaults to "vpnkit".
+# * DOCKERD_ROOTLESS_ROOTLESSKIT_MTU=NUM: the MTU value for the rootlesskit network driver. Defaults to 1500.
+# * DOCKERD_ROOTLESS_ROOTLESSKIT_DISABLE_HOST_LOOPBACK=(true|false): prohibit connections to 127.0.0.1 on the host (including via 10.0.2.2, in the case of slirp4netns). Defaults to "true".
+
 set -eu
 
 _tls_ensure_private() {
@@ -216,11 +222,18 @@ if [ "$1" = 'dockerd' ]; then
 			echo >&2 "error: attempting to run rootless dockerd but need 'user.max_user_namespaces' (/proc/sys/user/max_user_namespaces) set to a sufficiently large value"
 			exit 1
 		fi
+
+		DOCKERD_ROOTLESS_ROOTLESSKIT_DISABLE_HOST_LOOPBACK=${DOCKERD_ROOTLESS_ROOTLESSKIT_DISABLE_HOST_LOOPBACK:-true}
+		host_loopback="--disable-host-loopback"
+		if [ "$DOCKERD_ROOTLESS_ROOTLESSKIT_DISABLE_HOST_LOOPBACK" = "false" ]; then
+			host_loopback=""
+		fi
+
 		# TODO overlay support detection?
 		exec rootlesskit \
 			--net="${DOCKERD_ROOTLESS_ROOTLESSKIT_NET:-vpnkit}" \
 			--mtu="${DOCKERD_ROOTLESS_ROOTLESSKIT_MTU:-1500}" \
-			--disable-host-loopback \
+			$host_loopback \
 			--port-driver=builtin \
 			--copy-up=/etc \
 			--copy-up=/run \


Arkaniad:
> I don't love the idea of maintaining more lines/complexity here -- as I've noted over in [#543 (review)](https://github.com/docker-library/docker/pull/543#pullrequestreview-2977057367), I'm not actually convinced continuing to maintain the `rootless` variant (still labelled on Docker Hub as "experimental" 😅) makes sense.

Is there a way to configure the regular `dind` image to behave as the `dind-rootless` image does at runtime? If so, then possibly not - but most of the documentation I've found pertaining to rootless DinD setups specifically points out this image as necessary for it to work.

As @MadLittleMods mentioned, the key thing we are trying to accomplish is communication from guest containers to a process running in the host container to co-ordinate some testing. Investigation along [these lines](https://medium.com/@dsoplanner/investigating-docker-rootless-mode-networking-2cbf563df0cd) seems to indicate that for what we want to do, we'll need `slirp4netns`, which only seems to be built into the `rootless` variant. As well we'd like to enable loopback communication selectively in this configuration, which seems to be not possible without this MR's changes (unless there's a less-obvious way to toggle this switch in rootlesskit at runtime)

yosifkit:
This is already possible with `DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS`; last flag wins. You can see it complain about it being false in the output:
```console
$ docker run -it --rm --privileged -e DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS=--disable-host-loopback=false docker:29-dind-rootless
...
WARN[0000] [rootlesskit:parent] specifying --disable-host-loopback is highly recommended to prohibit connecting to 127.0.0.1:* on the host namespace (requires pasta, slirp4netns, or VPNKit)
```

MadLittleMods:
Ahh, perfect @yosifkit :ok_hand: Thank you!

For reference, I've confirmed it this way:

### Reproduction/confirmation instructions

Start the `dind` container (notice the warning which is complaining about `--disable-host-loopback` not being set which is what we want):
```shell
# Note: This is using `slirp4netns` by default but we can also define it explicitly with `DOCKERD_ROOTLESS_ROOTLESSKIT_NET`.
# Note: `slirp4netns` has its gateway/host hard-coded as `10.0.2.2` which we will utilize below.
docker run -it --rm \
	--privileged \
	--env DOCKERD_ROOTLESS_ROOTLESSKIT_FLAGS=--disable-host-loopback=false  \
	--env 'DOCKER_HOST=unix:///run/docker/docker.sock' \
	--volume /home/eric/Downloads/docker-rootless-scratch:/run/docker \
	--name dind-rootless \
	docker:29.5.2-dind-rootless
...
WARN[0000] [rootlesskit:parent] specifying --disable-host-loopback is highly recommended to prohibit connecting to 127.0.0.1:* on the host namespace (requires pasta, slirp4netns, or VPNKit)
```

We can doubly confirm that `--disable-host-loopback` is being set correctly by looking at the processes in the `dind` container. Notice how the `slirp4netns` process doesn't have `--disable-host-loopback`

```shell
$ docker exec -it dind-rootless sh

# We're piping to `cat` so the command output isn't cut-off because our terminal isn't wide-enough
/ $ ps aux | cat
PID   USER     TIME  COMMAND
    1 rootless  0:00 rootlesskit --net=slirp4netns --mtu=1500 --disable-host-loopback --port-driver=builtin --copy-up=/etc --copy-up=/run --disable-host-loopback=false -p 0.0.0.0:2376:2376/tcp docker-init -- dockerd --host=unix:///run/docker/docker.sock --host=tcp://0.0.0.0:2376 --tlsverify --tlscacert /certs/server/ca.pem --tlscert /certs/server/cert.pem --tlskey /certs/server/key.pem
   76 rootless  0:00 /proc/self/exe --net=slirp4netns --mtu=1500 --disable-host-loopback --port-driver=builtin --copy-up=/etc --copy-up=/run --disable-host-loopback=false -p 0.0.0.0:2376:2376/tcp docker-init -- dockerd --host=unix:///run/docker/docker.sock --host=tcp://0.0.0.0:2376 --tlsverify --tlscacert /certs/server/ca.pem --tlscert /certs/server/cert.pem --tlskey /certs/server/key.pem
   95 rootless  0:00 slirp4netns --mtu 1500 -r 3 76 tap0
  104 rootless  0:00 docker-init -- dockerd --host=unix:///run/docker/docker.sock --host=tcp://0.0.0.0:2376 --tlsverify --tlscacert /certs/server/ca.pem --tlscert /certs/server/cert.pem --tlskey /certs/server/key.pem
  107 rootless  0:00 dockerd --host=unix:///run/docker/docker.sock --host=tcp://0.0.0.0:2376 --tlsverify --tlscacert /certs/server/ca.pem --tlscert /certs/server/cert.pem --tlskey /certs/server/key.pem
  124 rootless  0:00 containerd --config /run/user/1000/docker/containerd/containerd.toml
  317 rootless  0:00 [containerd-shim]
  399 rootless  0:00 sh
  408 rootless  0:00 ps aux
  409 rootless  0:00 cat
```

We want to sanity check that we can actually contact an HTTP server on the host from within a container we spawn. Since it's difficult to install anything in the `dind` container itself, we'll use a separate `runner` container and share the Docker socket. 

Start a `runner` container and try contacting the server on the runner host from within a container:
```shell
# This is a quick dirty hack. Alternatively, you can specify the same user when running
# both the `dind` and `runner` container (`--user 1001:1001`)
chmod 777 /home/eric/Downloads/docker-rootless-scratch/docker.sock

# Using `--network container:dind-rootless` so we share the same network
# as the `dind` container so that any server we start in the `runner` container
# is accessible from the `dind` container. In a Kubernetes environment, we get
# this for free since all containers in a pod share the same network.
docker run -it --rm \
	--privileged \
	--env 'DOCKER_HOST=unix:///run/docker/docker.sock' \
	--volume /home/eric/Downloads/docker-rootless-scratch:/run/docker \
	--network container:dind-rootless \
	--name runner  \
	ghcr.io/actions/actions-runner:latest /bin/bash

# Start a server on the runner host
$> python3 -m http.server &
# Try contacting the server on the runner host from within a container
#
# In a more real setup, you'd use `--add-host host.docker.internal:host-gateway` and configure `host-gateway-ip` to be `10.0.2.2`,
# see https://docs.docker.com/reference/cli/dockerd/#configure-host-gateway-ip
$> docker run --add-host host.docker.internal:10.0.2.2 busybox wget host.docker.internal:8000
Unable to find image 'busybox:latest' locally
latest: Pulling from library/busybox
b05093807bb0: Pull complete
7270b3e1860c: Download complete
Digest: sha256:fd8d9aa63ba2f0982b5304e1ee8d3b90a210bc1ffb5314d980eb6962f1a9715d
Status: Downloaded newer image for busybox:latest
Connecting to host.docker.internal:8000 (10.0.2.2:8000)
127.0.0.1 - - [03/Jun/2026 19:32:07] "GET / HTTP/1.1" 200 -
saving to 'index.html'
index.html           100% |********************************|   896  0:00:00 ETA
'index.html' saved

# If you want to inspect things further, here are some useful utils
$> apt update
# For the `ip` command
$> apt install -y iproute2
# For the `ping` command
$> apt install -y iputils-ping
```

PR #575, opened 2026-05-13
Add 29.5.0

yosifkit:
`vpnkit` is no longer included in the `docker-rootless-extras-*.tgz` and is no longer necessary for RootlessKit 3.0.0: https://github.com/moby/moby/pull/52319

Since it is no longer included, the automated update job cannot successfully build the `29.5.0` release:
```console
+ wget -O rootless.tgz https://download.docker.com/linux/static/test/x86_64/docker-rootless-extras-29.5.0.tgz
...
+ tar --extract --file rootless.tgz --strip-components 1 --directory /usr/local/bin/ docker-rootless-extras/rootlesskit docker-rootless-extras/vpnkit
tar: docker-rootless-extras/vpnkit: not found in archive
```

### DIFF
--- a/Dockerfile-dind-rootless.template
+++ b/Dockerfile-dind-rootless.template
@@ -3,7 +3,8 @@ FROM docker:{{ env.version }}-dind
 
 # busybox "ip" is insufficient:
 #   [rootlesskit:child ] error: executing [[ip tuntap add name tap0 mode tap] [ip link set tap0 address 02:50:00:00:00:01]]: exit status 1
-# slirp4netns can be selected as rootlesskit's net driver using "-e DOCKERD_ROOTLESS_ROOTLESSKIT_NET=slirp4netns"
+# slirp4netns is recommended as rootlesskit's net driver
+# https://github.com/rootless-containers/rootlesskit/blob/be156df0429551f5c23dc61dd07201b33076ff1d/docs/network.md#--netslirp4netns-recommended
 RUN apk add --no-cache iproute2 fuse-overlayfs slirp4netns
 
 # "/run/user/UID" will be used by default as the value of XDG_RUNTIME_DIR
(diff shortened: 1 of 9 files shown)


yosifkit:
https://github.com/docker-library/docker/blob/132c91a93dc0d406ed5924189df0f9c8f2dfcd64/dockerd-entrypoint.sh#L221

~I think this may need adjusting; I'm looking into it.~ Updated to `slirp4netns`.

thaJeztah:
@AkihiroSuda could you have a quick peek if the changes for rootless are correct?

thaJeztah (commented):
LGTM

Dockerfile-dind-rootless.template
```diff
@@ -3,7 +3,8 @@ FROM docker:{{ env.version }}-dind
 
 # busybox "ip" is insufficient:
 #   [rootlesskit:child ] error: executing [[ip tuntap add name tap0 mode tap] [ip link set tap0 address 02:50:00:00:00:01]]: exit status 1
-# slirp4netns can be selected as rootlesskit's net driver using "-e DOCKERD_ROOTLESS_ROOTLESSKIT_NET=slirp4netns"
+# slirp4netns is recommended as rootlesskit's net driver
+# https://github.com/rootless-containers/rootlesskit/blob/be156df0429551f5c23dc61dd07201b33076ff1d/docs/network.md#--netslirp4netns-recommended
```
AkihiroSuda, line 7: (slirp4netns is no longer necessarily required, as gvisor-tap-vsock is now embedded in rootlesskit)

AkihiroSuda (commented):
LGTM, thanks

Dockerfile-dind-rootless.template
thaJeztah: In that case, should we simplify, and remove these bits ?

vvoland: Yeah I think we should get rid of the `slirp4netns is recommended as rootlesskit's net driver`, as its no longer recommended over the default?

AkihiroSuda: Yes, I suggest removing slirp4netns.

PR #578, opened 2026-07-29
Add back windows checksums

yosifkit:
With the `0.36.0` buildx release, the Windows builds moved to a separate checksums file (with darwin). Most of the change to `versions.sh` is whitespace.

ref https://github.com/docker/buildx/pull/3978

### DIFF
--- a/versions.sh
+++ b/versions.sh
@@ -63,7 +63,14 @@ buildxVersions="$(
 buildx=
 buildxVersion=
 for buildxVersion in $buildxVersions; do
-	if checksums="$(_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums.txt")"; then
+	# checksums.txt -> linux + *bsd checksums
+	# checksums-signed.txt -> windows + darwin checksums (see also https://github.com/docker/buildx/pull/3978)
+	# fail if one is missing
+	if \
+		checksums="$(_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums.txt")" \
+		&& checksums+=$'\n' \
+		&& checksums+="$(_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums-signed.txt")" \
+	; then
 		buildx="$(jq <<<"$checksums" -csR --arg version "$buildxVersion" '
 			rtrimstr("\n") | split("\n")
 			| map(
--- a/29/windows/windowsservercore-ltsc2022/Dockerfile
+++ b/29/windows/windowsservercore-ltsc2022/Dockerfile
@@ -41,8 +41,8 @@ RUN Write-Host ('Downloading {0} ...' -f $env:DOCKER_URL); \
 
 # https://github.com/docker-library/docker/issues/409#issuecomment-1462868414
 ENV DOCKER_BUILDX_VERSION 0.36.0
-ENV DOCKER_BUILDX_URL 
-ENV DOCKER_BUILDX_SHA256 
+ENV DOCKER_BUILDX_URL https://github.com/docker/buildx/releases/download/v0.36.0/buildx-v0.36.0.windows-amd64.exe
+ENV DOCKER_BUILDX_SHA256 ce84699d1d93a67d25888b9e6a717862a71b20305602f96d44770b1c12dfcdd8
 RUN $dir = ('{0}\docker\cli-plugins' -f $env:ProgramFiles); \
 	Write-Host ('Creating {0} ...' -f $dir); \
 	New-Item -ItemType Directory $dir -Force; \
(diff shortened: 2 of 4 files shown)


versions.sh
```diff
@@ -63,40 +63,46 @@ buildxVersions="$(
...
+	# checksums.txt -> linux + *bsd checksums
+	# checksums-signed.txt -> windows + darwin checksums (see also https://github.com/docker/buildx/pull/3978)
+	# two if statements to ensure that it fails if one is missing
 	if checksums="$(_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums.txt")"; then
-		buildx="$(jq <<<"$checksums" -csR --arg version "$buildxVersion" '
-			rtrimstr("\n") | split("\n")
-			| map(
-				split(" [ *]?"; "")
+		checksums+=$'\n' # add back trailing newline to separate the two checksums files
+		if checksums+="$(_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums-signed.txt")"; then
```
tianon: I don't love how deeply nested this makes everything -- why not capture both, with an `echo` in between if necessary?


```suggestion
	if checksums="$(
		set -Eeuo pipefail
		_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums.txt"
		echo
		_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums-signed.txt"
	)"; then
```

yosifkit: I wanted to make sure that either file not existing would fail so that we don't end up in the partial state again with missing windows urls. Even with `set -Eeuo pipefail` in the subshell, it doesn't fail early and still runs every command in it.

```bash
	if checksums="$(set -x -Eeuo pipefail 
		_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums.txt"
		_curl 'http://github.com/something-that-doesnt-exist'
		_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums-signed.txt")"; then
```
```console
++ _curl https://github.com/docker/buildx/releases/download/v0.36.0/checksums.txt
++ exec
++ local code
+++ curl --silent --location -o /dev/fd/42 --write-out '%{http_code}' https://github.com/docker/buildx/releases/download/v0.36.0/checksums.txt
++ code=200
++ case "$code" in
++ return 0
++ _curl http://github.com/something-that-doesnt-exist
++ exec
++ local code
+++ curl --silent --location -o /dev/fd/42 --write-out '%{http_code}' http://github.com/something-that-doesnt-exist
++ code=404
++ case "$code" in
++ return 1
++ _curl https://github.com/docker/buildx/releases/download/v0.36.0/checksums-signed.txt
++ exec
++ local code
+++ curl --silent --location -o /dev/fd/42 --write-out '%{http_code}' https://github.com/docker/buildx/releases/download/v0.36.0/checksums-signed.txt
++ code=200
++ case "$code" in
++ return 0
jq: error (at <stdin>:56): failed to parse os-arch from filename: Found8a75be22ecf40f633fe0a0199be48fa304c0446a0d7fa8a53d7c71b1d0093028

Comment (yosifkit) [reply versions.sh]: I think this would work for a single `if`.
```bash
if checksums="$(_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums.txt")"  \
	&& checksums+=$'\n'"$(_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums-signed.txt")"; then
```

Edit: I am unsure how to make nice whitespace with it.

tianon: Oh, that's probably the `set -e` propagation (or rather, lack thereof) inside our function -- I think it needs `|| return 1` on `curl` :thinking:

```console
$ bash -Eeuo pipefail -xc 'foo="$( echo yes; false; echo no )"; echo "$foo"'
++ echo yes
++ false
++ echo no
+ foo='yes
no'
+ echo 'yes
no'
yes
no

$ bash -Eeuo pipefail -xc 'foo="$( set -Eeuo pipefail; echo yes; false; echo no )"; echo "$foo"'
++ set -Eeuo pipefail
++ echo yes
++ false
+ foo=yes
$ # we fail on `foo=` because the whole line ends up returning `false`
```

yosifkit:
```console
$ bash -Eeuo pipefail -xc 'if foo="$( set -Eeuo pipefail; echo yes; false; echo no )"; then echo "$foo"; fi'
++ set -Eeuo pipefail
++ echo yes
++ false
++ echo no
+ foo='yes
no'
+ echo 'yes
no'
yes
no
```
😢 😭

tianon: oof; here's how I'd probably handle the whitespace in your suggested two-parter:

```suggestion
	if \
		checksums="$(_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums.txt")"  \
		&& checksums+=$'\n' \
		&& checksums+="$(_curl "https://github.com/docker/buildx/releases/download/v${buildxVersion}/checksums-signed.txt")" \
	; then
```

tianon (approved):

PR #584, opened 2026-09-04
dind: remove obsolete runtime dependencies

thaJeztah:
### dind: remove btrfs-progs dependency

The btrfs storage driver no longer requires the btrfs userspace
utilities at runtime.

Moby switched its btrfs implementation to use the kernel UAPI directly
in v23.0, removing its dependency on libbtrfs/btrfs-progs.

Containerd similarly switched its btrfs snapshotter to use the kernel
UAPI through containerd/btrfs/v2 in containerd 1.7, making kernel
headers a build-time dependency instead of requiring btrfs-progs at
runtime.

Remove btrfs-progs from the dind image, as neither the Moby btrfs
storage driver nor containerd's btrfs snapshotter requires it.

- https://github.com/moby/moby/commit/3208dcabdc8997340b255f5b880fef4e3f54580d
- https://github.com/containerd/containerd/commit/52f82acb7b3984fcb143af50c29a0ac0ceb5dc5d

### dind: remove e2fsprogs and xfsprogs dependencies

The e2fsprogs, e2fsprogs-extra, and xfsprogs packages were runtime
dependencies of the devicemapper storage driver, providing the tools
used to create and configure ext4 and XFS filesystems.

The devicemapper storage driver was disabled by default in Docker
Engine v23.0 and removed in v25.0. None of these filesystem utilities
are required by the remaining storage drivers.

Remove e2fsprogs, e2fsprogs-extra, and xfsprogs from the dind image.

- https://docs.docker.com/engine/deprecated/#device-mapper-storage-driver

### dind: remove duplicate git dependency

Git was originally added explicitly to the dind variants when the
separate git variant was folded into dind.

It was later added to the CLI variants as well, to support build and
buildx workflows that invoke git. As the dind image is based on the CLI
image, git is now inherited from its parent and no longer needs to be
installed again.

Remove the redundant git package from the dind dependencies.

- https://github.com/docker-library/docker/commit/485fefe743baed5a2dd9e5d22b685c14eda4c61e
- https://github.com/docker-library/docker/commit/b348a31fc955dd6500a4aa9537bbef51688a181e

### DIFF
--- a/Dockerfile-dind.template
+++ b/Dockerfile-dind.template
@@ -5,7 +5,6 @@ FROM docker:{{ env.version }}-cli
 # https://github.com/docker/docker-ce-packaging/blob/963aa02666035d4e268f33c63d7868d6cdd1d34c/deb/common/control#L28-L41
 RUN set -eux; \
 	apk add --no-cache \
-		git \
 		ip6tables \
 		iptables \
 		openssl \
--- a/Dockerfile-dind.template
+++ b/Dockerfile-dind.template
@@ -5,15 +5,12 @@ FROM docker:{{ env.version }}-cli
 # https://github.com/docker/docker-ce-packaging/blob/963aa02666035d4e268f33c63d7868d6cdd1d34c/deb/common/control#L28-L41
 RUN set -eux; \
 	apk add --no-cache \
-		e2fsprogs \
-		e2fsprogs-extra \
 		git \
 		ip6tables \
 		iptables \
 		openssl \
 		pigz \
 		shadow-uidmap \
-		xfsprogs \
 		xz \
 		zfs \
 	;
--- a/Dockerfile-dind.template
+++ b/Dockerfile-dind.template
@@ -5,7 +5,6 @@ FROM docker:{{ env.version }}-cli
 # https://github.com/docker/docker-ce-packaging/blob/963aa02666035d4e268f33c63d7868d6cdd1d34c/deb/common/control#L28-L41
 RUN set -eux; \
 	apk add --no-cache \
-		btrfs-progs \
 		e2fsprogs \
 		e2fsprogs-extra \
 		git \
--- a/Dockerfile-dind.template
+++ b/Dockerfile-dind.template
@@ -1,8 +1,9 @@
 {{ include "shared" -}}
 FROM docker:{{ env.version }}-cli
 
-# https://github.com/moby/moby/blob/0eecd59153c03ced5f5ddd79cc98f29e4d86daec/project/PACKAGERS.md#runtime-dependencies
-# https://github.com/docker/docker-ce-packaging/blob/963aa02666035d4e268f33c63d7868d6cdd1d34c/deb/common/control#L28-L41
+# Runtime dependencies:
+# - https://github.com/moby/moby/blob/864598f8c2677189e308eb4a3b3c0dd41fc06599/project/PACKAGERS.md#runtime-dependencies#runtime-dependencies
+# - https://github.com/docker/packaging/blob/0ba325d178593583f190dc91566214a85c3880ad/pkg/docker-engine/deb/control#L24-L36
 RUN set -eux; \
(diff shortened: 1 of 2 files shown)


thaJeztah:
- relates to https://github.com/docker-library/official-images/pull/22205

thaJeztah:
Goes from 528MB -> 519MB (unpacked + content)

```bash
docker image ls dind
dind:before     6f84404d1f9d        528MB          137MB
dind:lessdeps   7c41dc0bc569        519MB          134MB
```


cc @tianon @yosifkit @vvoland

Dockerfile-dind.template
```diff
@@ -1,20 +1,16 @@
...
-# https://github.com/docker/docker-ce-packaging/blob/963aa02666035d4e268f33c63d7868d6cdd1d34c/deb/common/control#L28-L41
+# Runtime dependencies:
+# - https://github.com/moby/moby/blob/864598f8c2677189e308eb4a3b3c0dd41fc06599/project/PACKAGERS.md#runtime-dependencies#runtime-dependencies
+# - https://github.com/docker/packaging/blob/0ba325d178593583f190dc91566214a85c3880ad/pkg/docker-engine/deb/control#L24-L36
 RUN set -eux; \
 	apk add --no-cache \
-		btrfs-progs \
-		e2fsprogs \
-		e2fsprogs-extra \
-		git \
```
vvoland, line 11: Doesn't Buildkit also depend on `git` for the git sources?

thaJeztah: It does, but the DIND image is based on the CLI image, which has it already installed

thaJeztah:
Also opened this one as draft (probably needs some looking what we want to keep);

- https://github.com/docker-library/docker/pull/585

Dockerfile-dind.template
vvoland: Oh, I should have read the commit body!

yosifkit (approved):

Dockerfile-dind.template
```diff
@@ -1,20 +1,16 @@
...
 		ip6tables \
 		iptables \
 		openssl \
 		pigz \
 		shadow-uidmap \
-		xfsprogs \
 		xz \
 		zfs \
```
tianon, line 15: Is ZFS really still actually supported?

tianon: Ah, I see you mention that over in https://github.com/docker-library/docker/pull/585, but IMO it has a similar shape/flavor to this PR to remove ZFS here.

thaJeztah: Yeah, I was considering removing it; for this PR I limited to dependencies that were not needed, so no behavior change. But honestly not sure if anyone would be using ZFS with this image.

thaJeztah: Let me know if you prefer removing it already, or in a follow up.

tianon: Yeah, I guess a follow-up would be good so we can let it simmer for a minute and see if anyone sees it and complains (although that will usually only happen after it's merged and released, but at least then there'll be a dedicated place to discuss / explain).

tianon:
https://github.com/docker-library/docker/pull/504#issuecomment-2201118180 lol, "later" has come :smile:

> For now, I've just added `git` to the CLI variants without removing it from the DinD variants because the later `apk add` should be a no-op, but I'm open to removing it and relying on the `FROM` relationship

thaJeztah:
> [#504 (comment)](https://github.com/docker-library/docker/pull/504#issuecomment-2201118180) lol, "later" has come 😄
> 
> > For now, I've just added `git` to the CLI variants without removing it from the DinD variants because the later `apk add` should be a no-op, but I'm open to removing it and relying on the `FROM` relationship

Ha! Missed that comment; I looked at the git history and found both commits (and did consider it was likely a no-op). 😅


I'll have a look at the follow up(s)

PR #586, opened 2026-09-08
dind: remove zfs dependency

thaJeztah:
- follow-up to https://github.com/docker-library/docker/pull/584#discussion_r3937818504

The zfs utilities were used by the zfs graph-driver. containerd provides an optional zfs-snapshotter, but it's not installed by default. Remove the dependency, as it's not used in the default configuration.

### DIFF
--- a/29/dind/Dockerfile
+++ b/29/dind/Dockerfile
@@ -17,7 +17,6 @@ RUN set -eux; \
 		pigz \
 		shadow-uidmap \
 		xz \
-		zfs \
 	;
 
 # dind might be used on systems where the nf_tables kernel module isn't available. In that case,
--- a/Dockerfile-dind.template
+++ b/Dockerfile-dind.template
@@ -12,7 +12,6 @@ RUN set -eux; \
 		pigz \
 		shadow-uidmap \
 		xz \
-		zfs \
 	;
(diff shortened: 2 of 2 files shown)


thaJeztah:
Size comparison;

```bash
docker image ls dind
dind:before     66a85fa1e49a        502MB          131MB
dind:ctr        b42924822491        471MB          122MB
```

thaJeztah:
Windows failure is unrelated;

```
failed to connect to the docker API at npipe:////./pipe/docker_engine; check if the path is correct and if the daemon is running: open //./pipe/docker_engine: The system cannot find the file specified.
Error: Process completed with exit code 1.
```

thaJeztah:
cc @tianon

tianon (approved):
LGTM

I think we should give this a little time to "soak" in case there are users who notice and can justify the continued inclusion of ZFS, but I also think we'll probably find those most effectively by merging and releasing this, so I think we should do that sometime next week. 👍

yosifkit:
Merged so that the change coincides with the `29.8.0` -> `29.8.1` bump. Users should pin to `29.8.0` (or install the package in a dependent image) if they need zfs.

